PR #1, opened 2016-02-04
Measure duration of experiments

RichardDalton:
Added the measurement of the experiments and added it to the ExperimentResult and the Measurement. 

Included a test to ensure this is populated correctly.


Haacked:
Ack! GitHub Selfie is not working so you'll have to settle for an oldie but goodie.

thumbs-up-oh-shit

Thanks! :sparkles:

PR #2, opened 2016-02-04
Added a Func to allow the user to pass in result comparison logic

RichardDalton:


Haacked:
I like the way this is heading. We should also add a check if the objects implement `IEquatable` and use that if no comparer is set. I'd also like to have a default comparer for objects that does a shallow property comparison.

RichardDalton:
I have now added the `IEquatable ` check and refactored the comparison code to a separate method.

Haacked:
Thanks! I'll try and look over this during the weekend.

Haacked:
> We should also add a check if the objects implement IEquatable and use that if no comparer is set.

So what I meant by this is we should check if the objects themselves implement `IEquatable`. I wasn't intending we add an equatable property.

Could you make just that change in a separate PR? I'd like to slowly iterate this to a better state. This PR is a lot to review and I'm not sure all of it is needed. I'd like to make one small change, try it out, and see what pain points we still have before adding more to it.

Make sense?

RichardDalton:
Thanks for the review. 

> I wasn't intending we add an equatable property.

Sorry, I might be getting confused. Where did I add an equatable property?

Haacked:
Whoops, I meant `IEqualityComparer` https://github.com/Haacked/Scientist.net/pull/2/files#diff-766a1d63ded5bb40834df3dfad293be1R15

akamud:
We were discussing some other alternatives to this implementation in #15.

Also, I'm planning on doing #16 as soon as this gets merged, I think this impacts directly on that issue.

RichardDalton:
Replaced this with #24

PR #3, opened 2016-02-04
Add NullGuard

Haacked:
This adds null checks to public surface areas. You can opt out using the
`[AllowNull]` attribute.


Haacked:
Self merge! YOLO!

PR #8, opened 2016-02-05
Added NSubstitute to remove side effect testing antipattern

darcythomas:
Now with less side effects.


Haacked:
Thanks, but I use NSubstitute these days.

darcythomas:
I will switch in NSubsitute.
Out of curiously why do you use NSubsitute?

Haacked:
Just like the syntax a bit better.
We've been using it with Octokit.net: https://github.com/octokit/octokit.net/blob/master/Octokit.Tests/Clients/AssigneesClientTests.cs

darcythomas:
Ok I am a convert. I like NSubsitute.
I have swapped out Moq with NSubsitute now

Haacked:
[selfie-1](https://github.com/thieman/github-selfies/)

Thanks!

PR #14, opened 2016-02-06
Align with ruby

joncloud:
Added new Observation and Result classes to match the Ruby counterparts for handling multiple experiment behaviors.
Updated InMemoryObservationPublisher to construct one completed task.
Altered IObservationPublisher to use new IResult interface for publishing all information about the observation.
Implemented the ability to compare exceptions using types and messages to match Ruby behaviors.
Minor updates to property naming to align with Ruby.


Haacked:
Ack, moar conflicts. Sorry!

joncloud:
@Haacked no worries.  Looks like I had an alternate implementation of 394171fd524f523bacd981f97bfece911b521f08 when I started aligning with ruby.

Should be all merged, just waiting appveyor.

joncloud:
Looks like the appveyor build was successful, and the merge path is good.

Haacked:
:cool: I'll need some time to review. I'm going on a ski trip for the weekend, so sorry for the delays.

src/Scientist/Internals/Experiment.cs
```diff
@@ -1,109 +1,32 @@
...
-                return new ExperimentResult(e, new TimeSpan(sw.ElapsedTicks));
-            }
-        }
+        public void Try(Func<T> candidate) { _candidate = () => Task.FromResult(candidate()); }
 
-        class ExperimentResult
-        {
-            public ExperimentResult(T result, TimeSpan duration)
+        internal ExperimentInstance<T> Build()
             {
```
Haacked: This is indented oddly. In fact, you don't need the braces if you make this an expression bodied member.

src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -0,0 +1,58 @@
...
+            : this(name, () => Task.FromResult(control()), () => Task.FromResult(candidate()))
+        {
+        }
+
+        public ExperimentInstance(string name, Func<Task<T>> control, Func<Task<T>> candidate)
+        {
+            _name = name;
+            _behaviors = new Dictionary<string, Func<Task<T>>>
+            {
+                { "control", control },
```
Haacked: Since you use "control" more than once, these strings should probably be constants.

Haacked: On second thought, why is this a dictionary in the first place?

src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -0,0 +1,58 @@
...
+            // TODO determine if experiments should be run.
+
+            // Randomize ordering...
+            var observations = new List<Observation<T>>();
+            foreach (string key in _behaviors.Keys.OrderBy(k => _random.Next()))
+            {
+                observations.Add(await Observation<T>.New(key, _behaviors[key]));
+            }
+
+            Observation<T> controlObservation = observations.FirstOrDefault(o => o.Name == name);
```
Haacked: Oh I see why it's a dictionary. Keep in mind the `OrderBy` doesn't affect the original dictionary. This could be a simple list and you just use reference equality here to compare it to the control item (which would be the first item in the list) rather than doing a string comparison here.

Added benefit, every item in the list after the first item would be a candidate (since we need to support more than one candidates eventually).

src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -0,0 +1,58 @@
...
+            Result<T> result = new Result<T>(_name, observations, controlObservation);
```
Haacked: Above two declarations should use `var`.

src/Scientist/IObservationPublisher.cs
```diff
@@ -8,10 +8,10 @@ namespace GitHub
     public interface IObservationPublisher
```
Haacked: Given the change below, we should probably rename this interface to `IResultPublisher` right?

src/Scientist/Observation.cs
```diff
@@ -1,49 +1,178 @@
...
 using System.Threading.Tasks;
 
 namespace GitHub
 {
     /// <summary>
-    /// Defines an observation of an experiment's execution.
+    /// Provides an interface for an observation of an experiment's execution.
     /// </summary>
-    public class Observation
+    public interface IObservation
```
Haacked: I don't think we need to introduce an interface here just yet. This type is pretty much just a DTO and it's easy to instantiate.

src/Scientist/Result.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System.Collections.Generic;
+using System.Linq;
+
+namespace GitHub
+{
+    /// <summary>
+    /// Provides an observed result.
+    /// </summary>
+    public interface IResult
```
Haacked: I don't think we need to introduce this interface. This is just an object that contains values (no real behavior) and it's easy to instantiate.

UnitTests/ScientistTests.cs
```diff
@@ -12,12 +12,41 @@ public class TheScientistClass
...
     {
         [Fact]
+        public void RunsBothBranchesOfTheExperimentAndMatchesExceptions()
+        {
+            bool candidateRan = false;
+            bool controlRan = false;
+
+            // We introduce side effects for testing. Don't do this in real life please.
+            // Do we do a deep comparison?
+            Func<int> control = () => { controlRan = true; throw new InvalidOperationException(); };
```
Haacked: In the recent unit tests, we use NSubstitute. Would be nice to update this to use that approach. You might need to merge master into this branch (or rebase against master and force push).

Haacked:
Ok, back to you.

src/Scientist/Internals/ExperimentInstance.cs
joncloud: This is to follow up with a change to allow multiple `Try` calls.  Alternately, it could be a `List` that has a property for storing the Name and Delegate.  What do you think?

src/Scientist/Observation.cs
joncloud: I built this and `IResult` to encapsulate the generic parameter on their implementations to keep the publishing method non-generic.  Here are a couple of thoughts:

* Update the publishing interface to be `Task Publish<T>(Result<T> result)`
* Provide a mechanism for scoping publishing to the specific type to allow `Task Publish(Result<T> result)` where the generic is defined on the class itself.
* Keep the interface to encapsulate the generic, and the publishing interface the same `Task Publish(IResult result)`.

Let me know what you think; i'm open to any of the avenues.

src/Scientist/Internals/Experiment.cs
joncloud: Most definitely - I updated all of the methods to follow that pattern.

src/Scientist/Internals/ExperimentInstance.cs
joncloud: After seeing the following comment I updated this to be a `List`.

src/Scientist/Internals/ExperimentInstance.cs
joncloud: Done

src/Scientist/IObservationPublisher.cs
joncloud: Yeah I also updated the default implementation of the interface to be `InMemoryResultPublisher`.

UnitTests/ScientistTests.cs
joncloud: All fixed up.  Definitely cleaner.

joncloud:
The only thing that should be outstanding is going to be the `IObservation` and `IResult` interfaces.  Let me know what you think about the publishing side along with the generic argument, and I can provide an implementation that'll suit our needs.

Have fun on the ski trip!

src/Scientist/Observation.cs
Haacked: Ah I see!

I think the problem with the interfaces for publishing is we actually lose static type information that might be useful to an implementation of the publisher. So I tend to lean towards:

> Update the publishing interface to be `Task Publish<T>(Result<T> result)`

That way a single instance of a publisher can handle all publishing requests.

src/Scientist/Internals/ExperimentInstance.cs
Haacked: :+1:

joncloud:
I submitted an additional update to account for `IResultPublisher.Publish<T>`.  Let me know what you think.

Haacked:

Haacked:
Nicely done! Thank you very much! :sparkles:

Issue #15, opened 2016-02-07
Support custom comparators


akamud:
Similar to Ruby's [original code](https://github.com/github/scientist/blob/master/lib/scientist/experiment.rb#L168).

I'm willing to implement this, but I would like to discuss the preferred implementation. I see a few ways to do this in C#:

### 1. Use an optional parameter
We could make `Science` have an optional parameter `Func<T, T, bool>` that would be called at the end of the experiment if it is not null.  
I implemented this in my fork, you can see the changes needed [here](https://github.com/akamud/Scientist.net/commit/e4fc4eebed4072d532b43798e30b33b4e176226b).

### 2. Add a property in Scientist class
Very similar to # 1, but instead of passing an optional parameter you would make Scientist generic (`Scientist<T>`), having a property `Func<T, T, bool> Comparer` that could be set like an `ObservationPublisher`.  
This would change the way the lib is used right now, as a type parameter would have to be passed in `Scientist` class but no parameter would be needed in the `Science` method anymore.

Old way:
```C#
var result = Scientist.Science<int>("success", experiment =>
...
```
New way
```C#
var result = Scientist<int>.Science("success", experiment =>
...
```
From my point of view this would make the code a little less clean, as the type would have to be used anytime the Scientist object is referenced.

### 3. Change `Experiment<T>` interface
Ruby's implementation of Comparator is inside the `Experiment` class, doing that in C# would make implementing the `Compare` method required, even for straightforward value types' comparison.  
Maybe we could provide an abstract `Experiment` class that would be easier to implement (default `Compare` virtual method using default C# comparators?) allowing the user to override when needed?

----

What do you think? Does any of these make sense? Is there a simpler more idiomatic way of doing this in C#?

giulianob:
I would opt for option #3. You can just have the default implementation in ExperimentBuilder perform a simple Equal but allow the user to set a different comparison function. It seems that would match the Ruby API the best.

darcythomas:
So I made a thing https://github.com/Haacked/Scientist.net/pull/26

When you setup a test you optionally can pass in a ` Func<T, T, bool>` or a IEqualityComparer<T> 
if neither are passed in the the default rules are applied

Specifically it checks like this: 

``` cs

        /// <summary>
        /// Checks if two ExperimentResults are equal
        /// </summary>
        /// <param name="controlResult">Control ExperimentResult</param>
        /// <param name="candidateResult">Candidate ExperimentResult</param>
        /// <returns>
[...]

PR #18, opened 2016-02-09
Should fix #16

### LINKED ISSUE #16: Support different return types for Control/Candidate methods
This implementation depends on decisions made for #15.

Basically the idea is that sometimes the candidate code returns a different, but relatable value. For example:

```C#
public int ControlMethod(int foo)
{
    return 1;
}

public bool CandidateMethod(int foo)
{
    return true;
}
```

The candidate method now returns a `bool`, and while `bool` and `int` are different, you could easily map `1 -> true` and `0 -> false`.

This could be implemented creating a new interface `Experiment<TControl, TCandidate>` that would allow the user to specify different types for each method. In this case, providing a `Compare` method could even be required for the interface.

What do you think?

jlandheer:
Added support for custom comparing of results.

### DIFF
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Runtime.CompilerServices;
 using System.Threading.Tasks;
 using GitHub;
 using GitHub.Internals;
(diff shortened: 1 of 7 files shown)


Haacked:
Ack! There's another PR already intending to add custom result comparisons.

jlandheer:
Ah, the power of looking before you code... The difference I see with @RichardDalton's code is the support for different return types (#15) which necessitates a custom comparer. His code looks much better though.

Haacked:
Ok, how would you like to proceed. Unfortunately, I merged a big change that created merge conflicts for this branch.

Haacked:
I'm going to close this one for now. Sorry about that. Feel free to look over the current implementation and let me know if you have thoughts to improve it.

jlandheer:
No problem.

PR #20, opened 2016-02-09
Fixed scientist.nuspec projectUrl value

PedroLamas:
I assume you changed the repo name after creating the nuspec file? :smile:


Haacked:
I certainly did. Thanks for the patch!

PR #21, opened 2016-02-10
Don't wait for observation to be published

darcythomas:


darcythomas:
This is all wrong. Killing. Will rethink.

joncloud:
What do you think about https://github.com/joncloud/Scientist.net/commit/cac40b89c25ca65fba1c6e40399c5321c47966b6?

PR #24, opened 2016-02-10
Simplified comparison update

RichardDalton:
With just the Func and the check for IEquatable.

Replacing [previous pull request](https://github.com/Haacked/Scientist.net/pull/2)

### DIFF
--- a/UnitTests/ComplexResult.cs
+++ b/UnitTests/ComplexResult.cs
@@ -1,0 +1,15 @@
+﻿using System;
+
+public class ComplexResult : IEquatable<ComplexResult>
+{
+    public int Count { get; set; }
+    public string Name { get; set; }
+
+    public bool Equals(ComplexResult other)
+    {
+        if (other == null)
+            return false;
+
+        return Count == other.Count && Name == other.Name;
+    }
+}
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Linq;
+using System.Runtime.Remoting.Messaging;
 using System.Threading.Tasks;
 using GitHub;
 using GitHub.Internals;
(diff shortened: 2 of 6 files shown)


UnitTests/ComplexResult.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+public class ComplexResult : IEquatable<ComplexResult>
```
Haacked: Ha! I kept wondering why this class is here and then I realized, this is for the unit tests! Silly me.

UnitTests/ScientistTests.cs
```diff
@@ -121,5 +122,103 @@ public void AnExceptionReportsDuration()
...
+
+        [Fact]
+        public void RunsBothBranchesOfTheExperimentWithResultComparisonSetAndReportsSuccess()
+        {
+            bool candidateRan = false;
+            bool controlRan = false;
+
+            // We introduce side effects for testing. Don't do this in real life please.
+            // Do we do a deep comparison?
+            Func<ComplexResult> control = () => { controlRan = true; return new ComplexResult {Count = 10, Name = "Tester"}; };
```
Haacked: Now that we are using NSubstitute, want to take a crack at converting this test to use that? Check out the other tests for examples.

UnitTests/ComplexResult.cs
RichardDalton: Maybe I should have named it better. Damn those really hard problems!

src/Scientist/Internals/Experiment.cs
```diff
@@ -30,6 +31,14 @@ public ExperimentInstance(string name, Func<Task<T>> control, Func<Task<T>> cand
             _candidate = candidate;
         }
 
+        public ExperimentInstance(string name, Func<Task<T>> control, Func<Task<T>> candidate, Func<T, T, bool> resultComparison)
```
Haacked: Seems like `resultComparer` would be a better name. To me, "comparison" sounds like the result of comparing. Thoughts?

src/Scientist/Internals/Experiment.cs
```diff
@@ -66,13 +72,30 @@ public ExperimentInstance(string name, Func<Task<T>> control, Func<Task<T>> cand
             return controlResult.Result;
         }
 
+        private bool CompareResults(ExperimentResult controlResult, ExperimentResult candidateResult)
+        {
+            if (_resultComparison != null)
+            {
+                return _resultComparison(controlResult.Result, candidateResult.Result);
```
Haacked: As I think about this more, I worry taht people supplying a comparer function will forget to handle null checks. In fact, why should we make them?

If exactly one of the results is null and the other is not, then we know to return false.
If both results are null, we return true.

If both results are not null, we then continue and use the result comparers etc.

That way, developers who pass in a `resultComparison` don't need to worry about checking for null. The tricky part is that `T` is not constrained to be a reference type. So we'd probably need to cast to `object` and then do the null checks. That leads to boxing in the case of value types, which sucks, but I think it makes the API easier to use. Thoughts?

src/Scientist/Internals/Experiment.cs
```diff
@@ -14,6 +14,7 @@ internal class ExperimentInstance<T>
 
         readonly Func<Task<T>> _control;
         readonly Func<Task<T>> _candidate;
+        private readonly Func<T, T, bool> _resultComparison;
```
Haacked: Mind removing `private` to adhere to the current convention?

Haacked:
:pineapple: Done with my review. Some comments for you.

UnitTests/ComplexResult.cs
Haacked: Nah, it's good. I just needed to look at where it lives.

RichardDalton:
Cheers, will make those changes now.

RichardDalton:
Changes done. In the CompareResults method I have moved the null checks to the top so any consumers do not need to worry about null checking the objects that are passed to their comparison code.

Haacked:
I'll fix up these conflicts as I want to make some slight changes.

PR #25, opened 2016-02-10
Add experiments with different return types, fixes #16

### LINKED ISSUE #16: Support different return types for Control/Candidate methods

akamud:
This is intended to allow scenarios described in #16.

I'll rebase/refactor and add unit tests when #24 is merged, I'm just doing a PR so you can review the code suggested.

### DIFF
--- a/src/Scientist/IExperiment.cs
+++ b/src/Scientist/IExperiment.cs
@@ -7,6 +7,20 @@ namespace GitHub.Internals
     /// Provides an interface for defining a synchronous experiment.
     /// </summary>
     /// <typeparam name="T">The return result for the experiment.</typeparam>
+    public interface IExperiment<TControl, TCandidate>
+    {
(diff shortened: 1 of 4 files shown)


akamud:
Needs refactoring since #24 is now merged.

PR #27, opened 2016-02-12
Fire forget  publishing of observations

darcythomas:
Publishing of observations was essentially synchronous (as it was the last `await` before returning from an experiment). 

I have refactored the ` Scientist.PublishObservation()`  to use a, thread safe, `BlockingCollection` which is not `async` (not that it matters in this case). Blocking collections allow fast, thread safe, adds; which a consumer then blocks reading from until there is an item to process at the other end.

This did mean that using the `Scientist.ObservationPublisher` property stopped work properly. (but Ihave it working now) As the Experiment may have ended before the Observation had been published (which messed up a bunch of unit tests). This lead to race conditions where one unit test would be Asserting against another tests Observations.

I then implemented an `IQbservable<Observation>` property `Observations` this allows us to have multiple Observation subscribers with filtering! *Snazzy* Which then pipes back into`Scientist.ObservationPublisher`

Finally to allow unit-tests to find their own observations, I added a `CallingMethodName` property to the `Observation` Class  and infrastructure to automatically populate the method name of where an experiment was run from. That with the `TestHelper.ObservationsGeneratedInThisMethod()` a unit test's correct observations can be found.

Example:

``` cs

        [Fact]
        public async Task AllowsReturningNullFromControlOrTest()
        {
            var result = Scientist.Science<object>("failure", experiment =>
                {
                    experiment.Use(() => null);
                    experiment.Try(() => null);
                });

            Assert.Null(result);

            var observation = await TestHelper.ObservationsGeneratedInThisMethod().FirstAsync();

            Assert.True(observation.Success);
        }


```

### DIFF
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -1,7 +1,5 @@
 ﻿using System;
-using System.Diagnostics;
 using System.Linq;
-using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using GitHub;
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -134,14 +134,5 @@ public class TheScientistClass
             Assert.True(observation.ControlDuration.Ticks > 0);
             Assert.True(observation.CandidateDuration.Ticks > 0);
         }
-
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        public string GetCurrentMethod()
-        {
-            StackTrace st = new StackTrace();
-            StackFrame sf = st.GetFrame(1);
-
-            return sf.GetMethod().Name;
-        }
     }
 }
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -1,5 +1,8 @@
 ﻿using System;
+using System.Diagnostics;
 using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
 using System.Threading.Tasks;
 using GitHub;
 using GitHub.Internals;
@@ -7,17 +10,19 @@ using NSubstitute;
 using UnitTests;
 using Xunit;
 
+using System.Reactive.Linq;
+
 public class TheScientistClass
 {
     public class TheScienceMethod
     {
         [Fact]
-        public void RunsBothBranchesOfTheExperimentAndReportsSuccess()
+        public async void RunsBothBranchesOfTheExperimentAndReportsSuccess()
         {
-            var mock = Substitute.For< IControlCandidate<int>>();
+            var mock = Substitute.For<IControlCandidate<int>>();
             mock.Control().Returns(42);
             mock.Candidate().Returns(42);
-           
+
 
             var result = Scientist.Science<int>("success", experiment =>
             {
@@ -28,7 +33,9 @@ public class TheScientistClass
             Assert.Equal(42, result);
             mock.Received().Control();
             mock.Received().Candidate();
-            Assert.True(((InMemoryObservationPublisher)Scientist.ObservationPublisher).Observations.First(m => m.Name == "success").Success);
+
+            var observation = await TestHelper.ObservationsGeneratedInThisMethod().FirstAsync();
+            Assert.True(observation.Success);
         }
 
         [Fact]
@@ -37,9 +44,6 @@ public class TheScientistClass
             var mock = Substitute.For<IControlCandidateTask<int>>();
             mock.Control().Returns(Task.FromResult(42));
             mock.Candidate().Returns(Task.FromResult(43));
-           
-
-
 
             var result = await Scientist.ScienceAsync<int>("failure", experiment =>
             {
@@ -50,20 +54,25 @@ public class TheScientistClass
             Assert.Equal(42, result);
             await mock.Received().Control();
             await mock.Received().Candidate();
-            Assert.False(TestHelper.Observation.First(m => m.Name == "failure").Success);
+
+            var observation = await TestHelper.ObservationsGeneratedInThisMethod().FirstAsync();
+            Assert.False(observation.Success);
         }
 
         [Fact]
-        public void AllowsReturningNullFromControlOrTest()
+        public async void AllowsReturningNullFromControlOrTest()
         {
             var result = Scientist.Science<object>("failure", experiment =>
-            {
-                experiment.Use(() => null);
-                experiment.Try(() => null);
-            });
+                {
+                    experiment.Use(() => null);
+                    experiment.Try(() => null);
+                });
 
             Assert.Null(result);
-            Assert.True(TestHelper.Observation.First(m => m.Name == "failure").Success);
+
+            var observation = await TestHelper.ObservationsGeneratedInThisMethod().FirstAsync();
+
+            Assert.True(observation.Success);
         }
 
         [Fact]
@@ -77,7 +86,7 @@ public class TheScientistClass
         }
 
         [Fact]
-        public void RunsBothBranchesOfTheExperimentAndReportsSuccessWithDurations()
+        public async void RunsBothBranchesOfTheExperimentAndReportsSuccessWithDurations()
         {
             var mock = Substitute.For<IControlCandidate<int>>();
             mock.Control().Returns(42);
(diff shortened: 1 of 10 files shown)


UnitTests/ScientistTests.cs
```diff
@@ -28,7 +33,9 @@ public void RunsBothBranchesOfTheExperimentAndReportsSuccess()
             Assert.Equal(42, result);
             mock.Received().Control();
             mock.Received().Candidate();
-            Assert.True(((InMemoryObservationPublisher)Scientist.ObservationPublisher).Observations.First(m => m.Name == "success").Success);
+
+            var observation = await TestHelper.ObservationsGeneratedInThisMethod().FirstAsync();
```
darcythomas: Thread safe way of getting a unit test observations

UnitTests/TestHelper.cs
```diff
@@ -6,4 +9,10 @@ public static class TestHelper
 {
     public static IEnumerable<Observation> Observation =>
         ((InMemoryObservationPublisher)Scientist.ObservationPublisher).Observations;
+
+    [MethodImpl(MethodImplOptions.NoInlining)]
+    public static IQbservable<Observation> ObservationsGeneratedInThisMethod([CallerMemberName] string callingMethodName = "")
```
darcythomas: Helper function to find  observations created in the same method which this function was called.

UnitTests/UnitTests.csproj
```diff
@@ -39,6 +39,26 @@
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
+    <Reference Include="System.Reactive.Core, Version=2.2.5.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
```
darcythomas: Adding Rx

src/Scientist/Internals/Experiment.cs
```diff
@@ -11,27 +13,60 @@ namespace GitHub.Internals
...
         static Random _random = new Random(DateTimeOffset.UtcNow.Millisecond);
-
+      
         readonly Func<Task<T>> _control;
         readonly Func<Task<T>> _candidate;
         readonly string _name;
+        private readonly string _callingmethodName;
 
-        public ExperimentInstance(string name, Func<T> control, Func<T> candidate)
+        public ExperimentInstance(string name, Func<T> control, Func<T> candidate, String callingmethodName = "")
```
darcythomas: Need to pass along the calling method name (eventually to an observation)

src/Scientist/Internals/Experiment.cs
```diff
@@ -11,27 +13,60 @@ namespace GitHub.Internals
...
         }
 
+     
+
         public async Task<T> Run()
         {
+            ExperimentResult candidateResult;
+            ExperimentResult controlResult;
+
+            Tuple<ExperimentResult, ExperimentResult> result = await RunExperiments();
```
darcythomas: Pushed the in to a method.

src/Scientist/Internals/Experiment.cs
```diff
@@ -47,24 +82,13 @@ public ExperimentInstance(string name, Func<Task<T>> control, Func<Task<T>> cand
...
-                || controlResult.Result != null && controlResult.Result.Equals(candidateResult.Result)
-                || controlResult.Result == null && candidateResult.Result != null;
-
-            // TODO: Get that duration!
-            var observation = new Observation(_name, success, controlResult.Duration, candidateResult.Duration);
+       
 
-            // TODO: Make this Fire and forget so we don't have to wait for this
-            // to complete before we return a result
-            await Scientist.ObservationPublisher.Publish(observation);
```
darcythomas: This is the one line I wanted to fix. Gah so much for one small functional change.

src/Scientist/Observation.cs
```diff
@@ -30,20 +32,25 @@ public Observation(string name, bool success, TimeSpan controlDuration, TimeSpan
         public bool Success { get; }
         
         /// <summary>
-        /// Gets the total duration for the controlled experiment that was executed through
+        /// Total duration of the controlled experiment.
```
darcythomas: Less ghostDoc like comments

src/Scientist/Observation.cs
```diff
@@ -30,20 +32,25 @@ public Observation(string name, bool success, TimeSpan controlDuration, TimeSpan
...
-        /// Gets the name of the experiment that was observed.
+        /// Name of the experiment that was observed.
         /// </summary>
         public string Name { get; }
-     }
+
+        /// <summary>
+        /// Name of the method where the experiment that was performed.
+        /// </summary>
+        public string CallingMethodName { get; }
```
darcythomas: New CallingMethodName Property

src/Scientist/Scientist.cs
```diff
@@ -11,15 +16,24 @@ namespace GitHub
     public static class Scientist
     {
         // TODO: Evaluate the distribution of Random and whether it's good enough.
-        static readonly Random _random = new Random(DateTimeOffset.UtcNow.Millisecond);                                
+        static readonly Random _random = new Random(DateTimeOffset.UtcNow.Millisecond);
+
+        private static readonly BlockingCollection<Observation> ObservationsToPublish = new BlockingCollection<Observation>();
+        private static readonly Lazy<IQbservable<Observation>> SetupObservations = new Lazy<IQbservable<Observation>>(
```
darcythomas: Lazy setup of BlockingColection to IQbservable piping

src/Scientist/Scientist.cs
```diff
@@ -27,14 +41,16 @@ public static IObservationPublisher ObservationPublisher
         /// <typeparam name="T">The return type of the experiment</typeparam>
         /// <param name="name">Name of the experiment</param>
         /// <param name="experiment">Experiment callback used to configure the experiment</param>
+       
         /// <returns>The value of the experiment's control function.</returns>
         [return: AllowNull]
-        public static T Science<T>(string name, Action<IExperiment<T>> experiment)
+        [MethodImpl(MethodImplOptions.NoInlining)] //So that we get the Source code CallerMemberName method name (may be lost when inlined in Release Mode).
+#pragma warning disable 1573
+        public static T Science<T>(string name, Action<IExperiment<T>> experiment, [CallerMemberName] string callingMethodName = "")
```
darcythomas: `[CallerMemberName] string callingMethodName = ""` get the method from where Scienceing this was called.

UnitTests/ScientistTests.cs
```diff
@@ -7,17 +8,19 @@
...
 
+using System.Reactive.Linq;
+
 public class TheScientistClass
 {
     public class TheScienceMethod
     {
         [Fact]
-        public void RunsBothBranchesOfTheExperimentAndReportsSuccess()
+        public async void RunsBothBranchesOfTheExperimentAndReportsSuccess()
```
joncloud: Should this return a `Task` here?  I know the article you had linked suggested that some testing frameworks may actually ignore this component.

darcythomas: Ha ha good catch. I don't think it should matter, since this is just a unit test.  I made it async as I needed to use await. Interesting to see how easy it is to accidentally let a async void slip in. I can see the value of that reflected unit test checking for them.

I will change to async Task when I get back to my dev PC.

darcythomas: I have updated to async Task now

Haacked:
I was hoping to start with a simpler solution that doesn't add a lot of dependencies to the library. For example, the Ruby library simply defers the responsibility to the publisher implementation. That might be a fair approach.

darcythomas:
I was hopeing sor a simpler solution too! 

The downside to offloading that to the publisher, is that an experiment will block until the publisher has completed. So a slow publisher reduces the performance of a system, with otherwise fast experiments.

That being said I could remove the RX stuff; leaving just the concurrent collections dependency. However the down side is we would need to put in a bunch of Thread.Sleep(1)  in all if the unit tests (looking at published results) to wait for the blocking collection consumer to pass on the result to the publisher instance. That and make sure one test doesn't steal another's published result. (I was having lots of issues with that while working thus all out)

Using the Rx IQbservable seemed a really good use case for this kind of data flow, and a superb way of sorting out the publishing race conditions.

joncloud:
Instead of using `Thread.Sleep(1)`, could you expose some kind of member on `Scientist`, or even `InMemoryResultPublisher` that exposes a `Task` which can be awaited while all observations are being published?

Haacked:
@joncloud I like that idea. Expose it on `InMemoryResultPublisher`.

darcythomas:
You still have a race condition problem. (The unit tests are run in parallel) .
You would need to replace the ViewBag with a List which each test could query until their result came through. 
It all seems a bit hacky / potentially memory leak / poor man IQbervable.

However I am just looking at the code on github (not in VS) and my have got myself into a mental rut. 
I will have a look a bit deeper this weekend.

joncloud:
I think unit tests may run in [parallel][].  Definitely an important piece to keep in mind though.
[parallel]: https://xunit.github.io/docs/running-tests-in-parallel.html

    By default, each test class is a unique test collection. Tests within the same test class will not run in parallel against each other.

Could the entire tracking be handled within the `InMemoryResultPublisher` class?  For example it handles batching of results, and then starts a new `Task` for submitting each of the results?

darcythomas:
My bad; the tests are not running in parallel, but since the Experiment exits before the BlockingCollection consumer publishes, you still can get race conditions. (Well I was at least)

Haacked:
Thanks for your contribution! We ended up going with a different approach in #83. Even so, that approach builds on ideas and work done in other PRs like this one which is much appreciated.

Issue #31, opened 2016-02-17
AnExceptionReportsDuration Test fails


jtreuting:
I'm submitting a fix shortly

jtreuting:
This was fixed with PR 32: https://github.com/Haacked/Scientist.net/pull/32 so closing

PR #32, opened 2016-02-17
Fixes issue #31.  Test was using the wrong experiment name and the re…

jtreuting:
…sults should not match since the candidate throws an exception.

### DIFF
--- a/UnitTests/ScientistTests.cs
+++ b/UnitTests/ScientistTests.cs
@@ -18,7 +18,7 @@ public class TheScientistClass
             var mock = Substitute.For<IControlCandidate<int>>();
             mock.Control().Returns(x => { throw new InvalidOperationException(); });
             mock.Candidate().Returns(x => { throw new InvalidOperationException(); });
-            const string experimentName = "RunsBothBranchesOfTheExperimentAndMatchesExceptions";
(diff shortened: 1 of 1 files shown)


UnitTests/ScientistTests.cs
```diff
@@ -138,8 +138,8 @@ public void AnExceptionReportsDuration()
             Assert.Equal(42, result);
             mock.Received().Control();
             mock.Received().Candidate();
-            Result<int> observedResult = TestHelper.Results<int>().First(m => m.ExperimentName == "success");
-            Assert.True(observedResult.Matched);
+            Result<int> observedResult = TestHelper.Results<int>().First(m => m.ExperimentName == "failure");
```
joncloud: Do you think it would be better to use the name of the method as the experiment name to prevent any kind of duplicates in the observed results?

jtreuting: Yeah I think it would be better.  I was following what was in place to get the test working as expected, but would be happy to change them.

jtreuting:
I agree with @joncloud that the experiment names should match the test names, so I went ahead and updated them for all the tests.

Haacked:
:heart:

PR #33, opened 2016-02-18
Added support for suppressing candidate calls conditionally.

joncloud:


Haacked:
Nice! Mind updating the README with an example of usage?

[selfie-3](https://github.com/thieman/github-selfies/)

PR #34, opened 2016-02-18
Fix #12

### LINKED ISSUE #12: Use PcgRandom instead of System.Random
[Line 13 of Scientist.cs](https://github.com/Haacked/Scientist.net/blob/dc8bbf5c1c7efad694335fe3bc6f00d0d5b8c925/src/Scientist/Scientist.cs#L13) says:

```
// TODO: Evaluate the distribution of Random and whether it's good enough.
```

PcgRandom ([NuGet](https://www.nuget.org/packages/PcgRandom), [GitHub](https://github.com/bgrainger/PcgRandom)) is a .NET port of the [PCG](http://www.pcg-random.org/) RNG algorithm. It's a drop-in replacement for `System.Random` that provides a high-quality PRNG.

If you're interested, I can submit a PR; just let me know. (I'd also fix the thread-safety bug in `Experiment.cs`; a `static` Random instance could be accessed simultaneously by multiple threads, but [it's not thread-safe](https://msdn.microsoft.com/en-us/library/system.random.aspx#ThreadSafety), and doing so can corrupt its internal state.)

joncloud:
Added a public Random property on Scientist, which is initialized to a better implementation and thread-safe calculation.


Haacked:
Nice! Let's go even simpler. Let's not make it a property of Scientist and just use this one private implementation.

joncloud:
Sounds good.  Should be good to go once appveyor picks it up.

joncloud:
Actually I think I misinterpreted you.  I'll need to move that to the `ExperimentInstance` class from `Scientist`.

Haacked:
Yeah, I think it should be a static private member of `ExperimentInstance`.

Haacked:
Have you been following the discussion in #12?

joncloud:
Yeah I should be able to address that no problem.

joncloud:
I reverted back to `Random`, and added a locking object to ensure thread-safety.  Also I integrated the .net core changes from master to ensure build compatibility.

src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -60,8 +61,14 @@ internal ExperimentInstance(string name, NamedBehavior control, NamedBehavior ca
             }
 
             // Randomize ordering...
+            NamedBehavior[] orderedBehaviors;
+            lock (_randomLock)
```
Haacked: Not a big deal, but since `_random` is a private instance and not shared anywhere, we could have just locked on `_random` itself.

Haacked:
[selfie-3](https://github.com/thieman/github-selfies/)

Issue #35, opened 2016-02-18
Support .NET Core and other platforms


Haacked:
I don't think Scientist.NET uses much that requires it to target .NET 4.5+ only. If someone wants to make this work on .NET core I'd love it.

MarcosMeli:
I will work in FileHelpers to support .Net Core, maybe I can start with Scientist that looks easier

Did you have a timeframe for this issue or I can look at it in the next week or two ?

Thanks

davezych:
@Haacked @MarcosMeli I'd like to start on this if that's cool. I've already been working on https://github.com/davezych/shience in .NET Core so I think it would make sense for me to migrate this. 

How cutting-edge would you like to be? If we support the current "stable" runtime of `rc1-update1` we can target `dnx451` and `dnxcore50`, but moving up to some nightlies will allow targeting Net Platform Standard which will get UWP, Windows Phone, DNX, Mono, Win8.1, etc with a single target. (You can view the table here: https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md#mapping-the-net-platform-standard-to-platforms).

Haacked:
@davezych fine with me if you've already done the work. :smile: The key thing for me is that the NuGet package still work for .NET 4.5+ projects. So, feel free to go to town on what you think is best here.

davezych:
@Haacked it looks like I accidentally left the csproj file in the main project. That should probably be removed to avoid confusion.

I'll open a new issue for scripting the nuget package.

Haacked:
> @Haacked it looks like I accidentally left the csproj file in the main project. That should probably be removed to avoid confusion.

Ok, I'll do that.

davezych:
@Haacked close?

jpsingleton:
I've been trying out the [latest package](https://www.nuget.org/packages/scientist/1.0.0-alpha5) as it looks from https://github.com/Haacked/Scientist.net/pull/37 like it should work in .NET Core, but I can't get it to build.

If I remove all references to the full .NET Framework so that the relevant section in my `project.json` looks like this:
```javascript
  "frameworks": {
    "dnxcore50": { }
  },
```
Then I get this build error:
```
NU1002	The dependency Scientist 1.0.0-alpha5 in project ... does not support framework DNXCore,Version=v5.0.
```
Does this only support the latest drops of .NET Core or is RC1 still supported?
Any ideas?

davezych:
[...]

PR #37, opened 2016-02-22
Move to .NET Core

davezych:
Move to .NET Core. Targets `rc1-update1` runtime, `net451` and `dotnet5.2` frameworks. 

Even though `dotnet5.2` targets .NET Framework 4.5, we must explicitly target `net451` on it's own so the tests will run. (`xunit` and `NSubstitute` aren't compatible with net standard, and there are dep problems targeting only net standard things right now. See (https://github.com/dotnet/corefx/issues/5540)). 

@Haacked I moved everything to Net core, meaning all `csproj`'s are gone in favor of `xproj`'s and `project.json`'s. Since we're targeting `dotnet5.2` anyone targeting 4.5+ can consume the lib, and I didn't see a need to keep the old project structure around. If for whatever reason you want it I can restore it.


Haacked:
Nice! Any ideas on getting to build on AppVeyor?

davezych:
@Haacked I didn't realize you had CI set up already! 

I added an `appveyor.yml` build script that builds the projects and runs the tests.

Haacked:
> I added an appveyor.yml build script that builds the projects and runs the tests.

SWEET! Let me try this out to make sure it still works in VS2015

Haacked:

Works great! Unfortunately, TestDriven.NET can't run these tests. But the built in test runner can.

Haacked:
Want to help us script the creation of our nuget package next?

jpsingleton:
@davezych @Haacked I can't get this to work on ASP.NET 5/Core RC1 Update 1, details in https://github.com/Haacked/Scientist.net/issues/35. Maybe I need an newer version of .NET Core?

PR #41, opened 2016-02-26
Build scripts

davezych:
Build script + nuget package creation. Used FAKE, which finds the local DNVM (available in appveyor), downloads the correct runtime version, builds the project, runs tests, generates package(s). 

Things to note:
- I had to remove NullGuard due to a dnx targeting bug. Targeting `dnx451` and `dotnet5.2` was causing build errors due to invalid references and missing mscorlib. David Fowler helped me figure out what was going on, but didn't tell me explicitly what the bug was, other than "it's a bug in dnx but dnx is dead so luckily I don't have to investigate."
 - In the future when we move to rc2 (and the dotnet cli) this should be resolved. 
- For versioning, I decided to pull the version # out of the release notes and edit the project.json.


davezych:
@Haacked I just realized the version numbering I have set up still won't allow to specify only one place. The goal was to support:
- Pulling the version number out of release notes
- Allowing builds to be versioned by build number, e.g. `1.0.0-*` being versioned as `1.0.0-{buildnumber}` by DNX. 

The way it's set up now will force us to set both the release notes and the project.json version (to add/remove the `-*` build number version). For "official" releases we'd have to remove that asterisk. 

Unless you have any good ideas, I don't think it's really necessary to have build number versions. I can remove that and we can leave version numbers to the readme.

Haacked:
@davezych Just a clarification, with SemVer, the part after the `-` is the prerelease part. So we have:

`MAJOR.MINOR.PATCH-prerelease-name#buildstamp`

For example,

`1.0.0-alpha#20150226`

So we could have the build stamp (aka build number) be supplied by DNX, but everything else by the README.

> I don't think it's really necessary to have build number versions. I can remove that and we can leave version numbers to the readme.

Having said all that, I'm fine with this plan. Let's keep it as simple as possible and only do things when we absolutely need them. :+1:

src/Scientist/project.json
```diff
@@ -1,11 +1,12 @@
-﻿{
-  "version": "1.0.0-*",
+{
+  "version": "0.1.0-*",
```
Haacked: We can't do this at this point because 1.0.0-alpha is already up on NuGet. One thing we _could_ do is:

`1.0.0-alpha*`

After all, 1.0.0-alpha* is considered a lower version than 1.0.0. So I plan to just keep incrementing that until we're ready for the big 1.0.

davezych:
@Haacked 

>Just a clarification, with SemVer, the part after the - is the prerelease part.

Right. I forgot specifically what it was called. My point was, though, that leaving the prerelease part in there and having it supplied by dnx means we still need to maintain both versions, otherwise all of our packages will be prerelease until we update the version in the project.json. "Updating the version" would just mean removing `-*` but still, it's another thing to do. 

> We can't do this at this point because 1.0.0-alpha is already up on NuGet.

Ah, right. I "reverted" the version in `project.json` and release notes to `1.0.0-alpha` and left off the dnx build number. As you said, we can add it if we need it.

PR #42, opened 2016-02-26
Update the version since 1.0.0-alpha3 is published

Haacked:


Haacked:
@davezych I messed up. 1.0.0-alpha3 is published. I'm just updating this to 1.0.0-alpha4 so when we push it, it'll update the version in the feed.

davezych:
:shipit:

Haacked:
I'm sure it'll build successfully. YOLOMERGE!

Issue #39, opened 2016-02-23
Provide a fluent interface


davezych:
Fluent interfaces are fairly standard in .NET APIs now. We set this up in Shience and it worked well. We chained our calls like this:

    var userHasAccess = Scientist.Science<bool>("experiment-name")
        .Use(() => IsCollaborator(user))
        .Try(() => HasAccess(user))
        .WithContext(new { userId = user.Id })
        .Execute(); //Or ExecuteAsync()

To keep consistent with the way the Scientist api works now, we could do something like:

    return Scientist.Science<bool>("experiment-name", experiment =>
        {
            experiment.Use(() => IsCollaborator(user))
                      .Try(() => HasAccess(user))
                      .WithContext(new { userId = user.Id })
        });

Makes for a cleaner api.

M-Zuber:
I would argue that in addition to cleanliness it makes it easier to understand what's going on

Haacked:
I'm a little divided on this. I think in some cases, fluent interfaces can be confusing. For example, take the following class:

``` csharp
public class Shape {
  public void Rotate() {...}
}
```

When you see this:

```csharp
var shape = new Shape();
shape.Rotate();
```

You rightly expect that `Rotate` mutates the instance of `Shape`.

But if you define Rotate like this:

``` csharp
public class Shape {
  public Shape Rotate() {...}
}
```

And saw this:

```csharp
var shape = new Shape();
var rotated = shape.Rotate();
```

Intuitively, I would expect that `shape` is the original `Shape` and `rotated` is the rotated shape. In other words, when I see a `void` method I expect the method to mutate the object. But if I see a method that returns the same type that the method belongs to, I expect the method to not mutate the object but return a new instance.

Haacked:
So to finish my thought, the fact that we mutate the `experiment` instance makes me think the API of `Experiment` should not be fluent.

Haacked:
And to continue my stream of consciousness, I've never been that much of a fan of fluent interfaces because I think they've been overused and often make code harder to understand in an effort to make it more readable.

That being said, one area that fluent interfaces often shine is in cases where you have a builder pattern. And as I think about this, Scientist seems to fit well with that pattern as we are building up an experiment.

I've tried to stay close to the Ruby API because I want the documentation for Scientist to be easily translatable to the version we have here. At the same time, I want to make sure we stay true to C# idioms. I feel what we have today strikes a nice balance.

[...]

PR #43, opened 2016-02-29
Fluent interface

davezych:
@Haacked here's a quick implementation of the Shience style fluent interface. I don't know if I like it better and worse than the existing API... It's worth noting that this current implementation mutates the `Experiment` object each time but we could easily make it immutable. 

I think it's a bit cleaner, but I also don't think it adds a whole lot over what's already there.


M-Zuber:
> I think it's a bit cleaner, but I also don't think it adds a whole lot over what's already there.

:point_up:

davezych:
Hey @Haacked I don't see a need for this, I think the existing interface is fine. I'll close it, can you close #39?

PR #45, opened 2016-02-29
M-M-M-MULTI-TRY

davezych:
Allow user to call `Try` multiple times. The doesn't have to set a name for the first try, but any additional calls will require using the overload with a name specified.


Haacked:
:metal:

PR #48, opened 2016-03-14
Experiment context

davezych:
Add context to an experiment. Takes a string and an object, adds to a dictionary and supplies a readonly dictionary with the result for publishing.

### DIFF
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -18,6 +18,7 @@ namespace GitHub.Internals
         Func<Task> _beforeRun;
         Func<Task<bool>> _runIf = _alwaysRun;
         readonly List<Func<T, T, Task<bool>>> _ignores = new List<Func<T, T, Task<bool>>>();
+        readonly Dictionary<string, object> _contexts = new Dictionary<string, object>();
 
         public Experiment(string name)
         {
@@ -79,8 +80,11 @@ namespace GitHub.Internals
         public void Ignore(Func<T, T, Task<bool>> block) => 
             _ignores.Add(block);
 
+        public void Context(string key, object data) =>
+            _contexts.Add(key, data);
+
         internal ExperimentInstance<T> Build() =>
-            new ExperimentInstance<T>(_name, _control, _candidates, _comparison, _beforeRun, _runIf, _ignores);
+            new ExperimentInstance<T>(_name, _control, _candidates, _comparison, _beforeRun, _runIf, _ignores, _contexts);
 
         public void Compare(Func<T, T, bool> comparison)
         {
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -89,17 +89,6 @@ namespace GitHub.Internals
             _contexts.Add(key, data);
         }
 
-        public bool TryAddContext(string key, dynamic data)
-        {
-            if (_contexts.ContainsKey(key))
-            {
-                return false;
-            }
-
-            _contexts.Add(key, data);
-            return true;
-        }
-
         internal ExperimentInstance<T> Build() =>
             new ExperimentInstance<T>(_name, _control, _candidates, _comparison, _beforeRun, _runIf, _ignores, _contexts);
 
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -18,7 +18,7 @@ namespace GitHub.Internals
         Func<Task> _beforeRun;
         Func<Task<bool>> _runIf = _alwaysRun;
         readonly List<Func<T, T, Task<bool>>> _ignores = new List<Func<T, T, Task<bool>>>();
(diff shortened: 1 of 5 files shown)


src/Scientist/Internals/Experiment.cs
```diff
@@ -18,6 +18,7 @@ internal class Experiment<T> : IExperiment<T>, IExperimentAsync<T>
         Func<Task> _beforeRun;
         Func<Task<bool>> _runIf = _alwaysRun;
         readonly List<Func<T, T, Task<bool>>> _ignores = new List<Func<T, T, Task<bool>>>();
+        readonly Dictionary<string, object> _contexts = new Dictionary<string, object>();
```
davezych: Upon thinking about it, would this be better as a `Dictionary<string, dynamic>`? Hopefully would mean less casting when publishing.

test/Scientist.Test/ScientistTests.cs
```diff
@@ -525,5 +525,72 @@ public void ExperimentIgnoredWithComplexIgnoreFunc()
...
+            {
+                e.Use(mock.Control);
+                e.Try(mock.Candidate);
+                e.Context("test", "data");
+            });
+
+            var publishResults = TestHelper.Results<int>().First(m => m.ExperimentName == experimentName);
+
+            Assert.Equal(42, result);
+            Assert.True(publishResults.Contexts.Any());
```
joncloud: Should this verify the contents of `Contexts` instead of just making sure something exists?

davezych: I didn't really think it was necessary, but it could... same for the following test I suppose too.

src/Scientist/Internals/Experiment.cs
```diff
@@ -79,8 +80,11 @@ public void Try(string name, Func<T> candidate)
         public void Ignore(Func<T, T, Task<bool>> block) => 
             _ignores.Add(block);
 
+        public void Context(string key, dynamic data) =>
+            _contexts.Add(key, data);
```
Haacked: This could throw if the key matches an existing key. Maybe we should follow the .NET convention and add a `bool TryAddContext(key, data)` method.

src/Scientist/Internals/Experiment.cs
```diff
@@ -68,19 +72,36 @@ public void Try(string name, Func<T> candidate)
...
-        public void Ignore(Func<T, T, Task<bool>> block) => 
+        public void Ignore(Func<T, T, Task<bool>> block) =>
             _ignores.Add(block);
 
+        public void AddContext(string key, dynamic data)
+        {
+            _contexts.Add(key, data);
+        }
+
+        public bool TryAddContext(string key, dynamic data)
```
davezych: @Haacked I added the `TryAddContext`, but I have since realized I'm not *entirely* sure I see the use case here. I highly doubt anyone will accept user input for this, which means all contexts will be dev added, which means duplicate keys can be handled at "code time".

src/Scientist/Internals/Experiment.cs
```diff
@@ -68,19 +72,36 @@ public void Try(string name, Func<T> candidate)
...
 
-        public void Ignore(Func<T, T, bool> block) => 
+        public void Ignore(Func<T, T, bool> block) =>
             _ignores.Add((con, can) => Task.FromResult(block(con, can)));
 
-        public void Ignore(Func<T, T, Task<bool>> block) => 
+        public void Ignore(Func<T, T, Task<bool>> block) =>
             _ignores.Add(block);
 
+        public void AddContext(string key, dynamic data)
```
davezych: I didn't see a need to catch/rethrow/throw-our-own here and let it bubble the `ArgumentException` up if a duplicate key is added.

src/Scientist/Internals/Experiment.cs
Haacked: Hmm, good point. The idea here is you wouldn't pass  an experiment around and keep adding context in other  places, thus adding a duplicate context is a developer error and thus should throw?

If so, I agree. :smile: In that case, forget what I said. Sorry for the churn here.

src/Scientist/Internals/Experiment.cs
Haacked: Agreed.

davezych:
Yoh @Haacked where do we stand here? And when did you want to push out `beta5`, and where do you want to get before a `1.0` release?

Haacked:
I was on vacation. This looks good. I'd like to push out a beta ASAP. I need some time to remind myself how. :stuck_out_tongue: 

As for 1.0, what are your thoughts?

davezych:
I'd say we're getting close. A brief review of the ruby lib shows that we're missing a few minor features ([handling errors](https://github.com/github/scientist#handling-errors), [raise on mismatch](https://github.com/github/scientist#testing), [cleaning](https://github.com/github/scientist#keeping-it-clean)) but other than that it looks like most things are done.

We should probably get those few things I mentioned in before 1.0 so it's a *true* port...

PR #56, opened 2016-05-02
Added the ability to track and manage exceptions during an experiment. Fixed #55.

### LINKED ISSUE #55: Support exception tracking during operations.
[Error handling](https://github.com/github/scientist#handling-errors)

joncloud:


test/Scientist.Test/ThrownTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            experiment.Compare((x, y) =>
+            {
+                throw ex;
+            });
+            experiment.Use(mock.Control);
+            experiment.Try(mock.Candidate);
+        });
+
+        Assert.Equal(expectedResult, result);
+        mock.Thrown(Operation.Compare, ex);
```
Haacked: Did you mean to verify that the `Thrown` method was called with these parameters on the mock?

src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -106,12 +129,37 @@ internal ExperimentInstance(string name, NamedBehavior control, IEnumerable<Name
...
         
+        /// <summary>
+        /// Determine whether or not the experiment should run.
+        /// </summary>
+        async Task<bool> ShouldExperimentRun()
+        {
+            try
+            {
+                // TODO Implement Enabled here.
+                return Behaviors.Count > 1 && await RunIfAllows();
```
Haacked: The count should be greater than one? Is that because the behaviors collection includes both the control and the candidate? If so, might be useful to comment that.

test/Scientist.Test/ThrownTests.cs
joncloud: Whoops; yeah that's what I meant.

src/Scientist/Internals/ExperimentInstance.cs
joncloud: Sure thing. This method was a port of the same function in the ruby framework.

joncloud:
@Haacked should be good to go.

Haacked:
Looks good to me. I haven't had a chance to try it out. @davezych do you have any feedback before I merge?

davezych:
I don't have time to pull it down and try it but it looks good on paper.

As a note for the future, the `ExperimentInstance` constructor is getting a bit unwieldy - there's what, 10 parameters there now? We might want to think about somehow cleaning that up sometime.

Haacked:
> As a note for the future, the ExperimentInstance constructor is getting a bit unwieldy - there's what, 10 parameters there now? We might want to think about somehow cleaning that up sometime.

Indeed.

Haacked:
Nice work @joncloud!

PR #58, opened 2016-05-03
Created settings class to consolidate constructor for ExperimentInstance<T>.

### LINKED ISSUE #57: Do something about the unwieldy ExperimentInstance constructor.
The constructor for  [`ExperimentInstance`](https://github.com/Haacked/Scientist.net/blob/master/src/Scientist/Internals/ExperimentInstance.cs#L28) has a lot of parameters. Because we want this to effectively be an read only instance, we need to supply all the parameters via the constructor.

Our options are:

1. Create a new `ExperimentSettings` class that has settable properties. Add a constructor overload to `ExperimentInstance` that takes a settings instance and sets all the properties accordingly.
2. Create a new `ExperimentBuilder` class that lets you set properties and returns a new instance of `ExperimentInstance` with those properties set via some `Create` method.
3. Add multiple ctor overloads to `ExperimentInstance`
4. Something I hadn't considered...

/cc @davezych who noted this problem

joncloud:
Fixed #57.

### DIFF
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -98,7 +98,19 @@ namespace GitHub.Internals
         }
 
         internal ExperimentInstance<T> Build() =>
-            new ExperimentInstance<T>(_name, _control, _candidates, _comparison, _beforeRun, _runIf, _ignores, _contexts, ThrowOnMismatches, _thrown);
+            new ExperimentInstance<T>(new ExperimentSettings<T>
+            {
+                BeforeRun = _beforeRun,
+                Candidates = _candidates,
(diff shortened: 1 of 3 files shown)


Haacked:
@davezych any thoughts on this?

davezych:
I like this better than any of the alternatives you mentioned. Multiple constructor overloads don't really make sense because this is only called internally/that just mucks up things even more, and I don't see much value in creating a builder over just instantiating/passing settings in. 

:+1:

Haacked:
Agreed. I forgot that the end-user never actually creates this.

Haacked:
Thanks @joncloud! This is nice. :)

Issue #51, opened 2016-04-18
[META] Prepare for 1.0 release


davezych:
I'd like to open a discussion about where we stand right now and where we need to get for an official 1.0 release. 
### Things that are in the ruby lib that we don't have:
- [~~Keeping it clean~~](https://github.com/github/scientist#keeping-it-clean) (implemented by joncloud in #67)
- [~~Enabled~~](https://github.com/github/scientist#ramping-up-experiments) (however this functionality can be done using `RunIf`) (implemented by joncloud in #60)
- [~~Raise on mismatch~~](https://github.com/github/scientist#testing) (for testing purposes) (implemented by joncloud in #53)
- [~~Error handling~~](https://github.com/github/scientist#handling-errors) (implemented by joncloud in #56)

In order to be an _official_ port, these all need to be implemented. 
### Documentation

The docs need to be updated. In my mind we should probably steal the ruby docs and modify the sample code with C# examples. @Haacked - from what I remember this was your intention as well, right? And I'm assuming it's okay if we "steal" the ruby docs, right?
### Testing

I have used Scientist in an app I'm developing and so far haven't run into any issues. This is good. However are we confident in the product and it's stability? I'm not sure how to answer that other than to use it more, and also to wait some time and see if we get any bug reports. 
### Official logo

We need an official logo for the nuget package. I can't really find a logo for the ruby lib that we could mimic and .net-ify. What do we want here?

joncloud:
I should have some time to help port some of the missing features early this week.

davezych:
@joncloud great! If I haven't created issues to track each of the features make sure you create one for whatever you work on so we know it's in the queue.

joncloud:
Will do sounds like a plan.

Haacked:
@davezych This sounds like a great plan! I'm sorry I've been slow to respond. I'm speaking at a conference in Peru and have just been swamped.

joncloud:
@davezych @Haacked what is your opinion on the implementation of [Enabled](https://github.com/github/scientist#ramping-up-experiments)?

It seems like the architecture of C# and Ruby differ enough that this becomes the same as `RunIf` in the C# implementation.  I prepared this implementation (1e1053aea35e3c299cfcc39e86f982f3006c5bb2), but it does not seem to offer much.

davezych:
Yea, `Enabled` is basically a specific case of `RunIf`. I can make a case for both having and not having it.

[...]

PR #60, opened 2016-05-04
Implemented the ability to enable/disable experiments.

joncloud:


Haacked:
Remind me again what the difference between `Enabled` and `RunIf` is? They both seem to take a lambda that returns a boolean.

joncloud:
Check out our [discussion](https://github.com/Haacked/Scientist.net/issues/51#issuecomment-216703531) on #51.

Haacked:
Ah cool, I'll continue the discussion here with the PR.

> Yea, `Enabled` is basically a specific case of `RunIf`.

In this code, it looks like it's exactly the same thing, not just a specific case. How is `RunIf` more generic than `Enabled`?

I haven't looked at the Ruby code. What's the semantic difference between `RunIf` and `Enabled` there? If the implementation is the same exact thing, I'd almost rather have `Enabled` and remove `RunIf` because I like the naming better.

joncloud:
It looks like `RunIf` matches the functionality provided in ruby, but `Enabled` is driven through what seems like derivation of some sorts.

RunIf Example

```ruby
class DashboardController
  include Scientist

  def dashboard_items
    science "dashboard-items" do |e|
      # only run this experiment for staff members
      e.run_if { current_user.staff? }
      # ...
  end
end
```

Enabled Example
```ruby
class MyExperiment < ActiveRecord::Base
  include Scientist::Experiment
  def enabled?
    percent_enabled > 0 && rand(100) < percent_enabled
  end
end
```

Haacked:
Ah, I get it now. So yeah, it seems like this is one case where our implementation is different from the Ruby implementation because of differences in the languages.

Ruby allows you to override the `new` method so it's effectively a factory. That allows them to nicely replace any creation of `Experiment` with a custom `MyExperiment` class and then override the `enabled` method.

So the semantic difference here appears to be that `Enabled` is a _global_ enabled, and `RunIf` is specific to a single experiment.

So with that in mind, I think there's two approaches we should consider.

1. Don't implement `Enabled`. At least not in this way.
2. Mimic the Ruby approach by allowing users to register their own `Experiment` type and create experiments via a factory method. Thus `Enabled` is a virtual method of the `Experiment` base class.

What do you think?

joncloud:
Since `Experiment` builds an `ExperimentInstance`, how would it look to provide the call to `Enabled`?  Would we still follow the pattern of providing a delegate to `ExperimentInstance`, but it would point to `this.Enabled`?

Haacked:
@joncloud sorry for the delayed response. It occurs to me that maybe the type that needs to be "registered" ala the Ruby approach is a custom implementation of `ExperimentInstance`, not `Experiment` given our current implementation.

I think the important point is that `Enabled` needs to be global. We don't absolutely have to follow the Ruby approach if we come up with a more idiosyncratic approach.

Unfortunately, I don't really have time to look at this deeply right now.

joncloud:
Ah - That provides a bit more clarity to the requirement.  Let me take a stab at it, and i'll post something up afterward.

joncloud:
@davezych @Haacked let me know what you think.

src/Scientist/Internals/Experiment.cs
```diff
@@ -18,15 +18,17 @@ internal class Experiment<T> : IExperiment<T>, IExperimentAsync<T>
...
         private Func<T, T, bool> _comparison = DefaultComparison;
         private Func<Task> _beforeRun;
+        private readonly ILaboratory _laboratory;
         private Action<Operation, Exception> _thrown = _alwaysThrow;
         private Func<Task<bool>> _runIf = _alwaysRun;
         private readonly List<Func<T, T, Task<bool>>> _ignores = new List<Func<T, T, Task<bool>>>();
         private readonly Dictionary<string, dynamic> _contexts = new Dictionary<string, dynamic>();
 
-        public Experiment(string name)
+        public Experiment(string name, ILaboratory laboratory)
```
Haacked: What do you think about making this parameter a `Func<Task<bool>>` (and perhaps a `Func<bool>` overload too) instead of introducing this interface. I imagine in most cases, when I want to disable a set of tests, I won't want to create a whole implementation. I'll just create a lambda expression.

joncloud: I think my idea was to encapsulate other things into this interface at a global level, but that could easily be done as needed when additional things are identified.

What about implementing it like this?

```csharp
class Scientist {
    internal static Func<Task<bool>> _enabled = () => Task.FromResult(true);

    public static void Enabled(Func<bool> enabled) { ... }
    public static void Enabled(Func<Task<bool>> enabled) { ... }

    public static T Science<T>(...) {
        var experiment = new Experiment<T>(name, _enabled);
    }
}
```

Haacked: Looks good to me!

I agree with not adding the interface till we know what those other things are.

README.md
```diff
@@ -48,15 +48,10 @@ public decimal GetUserStatistic(IUser user)
 }
 ```
 
-Control whether experiments run on a global level by defining a new laboratory.
+Control whether experiments run on a global level.
```
Haacked: Maybe something like "Provide a global condition that determines whether experiments run or not."

Then in the example, let's put a realistic expectation. For example, maybe you'd want to turn off experiments during peak hours.

joncloud: I think that'll help out quite a bit.

joncloud:
Whoops - closed this branch too early.

Haacked:
Thanks a lot! :sparkles: Nice work!

PR #61, opened 2016-05-26
Upgraded to .NET Core RC2.

### LINKED ISSUE #54: dnxcore50 not supported
According to @jpsingleton in #35 ([this comment](https://github.com/Haacked/Scientist.net/issues/35#issuecomment-213725339)), `dnxcore50` is not supported. He gets an error saying:

>NU1002  The dependency Scientist 1.0.0-alpha5 in project ... does not support framework DNXCore,Version=v5.0.

joncloud:
I think this may resolve #54, however the new release is still a little confusing to me.  I created a new local package, and installed it to an RC2 ConsoleApp with success.

@davezych do you have any experience with RC2?

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -5,8 +5,6 @@ open System
 open System.Collections.Generic
 open System.Text
 
-let mutable DnxHome = "[unknown]"
-
 let architecture = getBuildParamOrDefault "architecture" "x86"
 let runtime = getBuildParamOrDefault "runtime" "clr"
 let runtimeVersion = getBuildParamOrDefault "runtimeVersion" "1.0.0-rc1-update1"
@@ -47,21 +45,8 @@ let Run workingDirectory fileName args =
 
     ProcessResult.New code messages errors
 
-let GetHomeDirectory =
-    let result = Run currentDirectory "cmd" "/c \"echo %USERPROFILE%\""
-    result.Messages.[0]
-
-let GetDnvmHome =
-    let homeDirectory =
-        if buildServer = BuildServer.AppVeyor
-            then "C:\\Program Files\\Microsoft DNX\\Dnvm\\"
-            else (GetHomeDirectory + "\\.dnx\\bin\\")
-
-    homeDirectory
-
-let GetDnxHome =
-    let homeDirectory = GetHomeDirectory
-    homeDirectory + ("\\.dnx\\runtimes\\dnx-" + runtime + "-win-" + architecture + "." + runtimeVersion + "\\bin\\")
+let dotnetHome = "C:\\Program Files\\dotnet\\"
+let dotnetExe = dotnetHome + "dotnet.exe"
 
 let UpdateProjectJson projectJson =
     let fullJsonPath = (__SOURCE_DIRECTORY__ + projectJson)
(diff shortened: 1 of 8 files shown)


joncloud:
Just noting: I missed the build scripts. If this is the right direction, then I can address that too.

jpsingleton:
At a glance, this looks like the right direction. But I don't have the time to test it just now.

It might be worth looking at how [Dapper did this upgrade](https://github.com/StackExchange/dapper-dot-net/pull/518). Although they had to tweak a few things after that, so worth looking at the commits in master.

joncloud:
I updated the dotnet target to be netstandard1.1, and updated the build scripts to use the new dotnet cli.  One thing stood out though in the build output is that the dotnet pack command seemed to name the artifacts based upon the old version (not the one in the current project.json.  I am unsure why it did that, but I may be missing something.

davezych:
@joncloud sorry, been MIA for a whole bunch of different reasons this whole month. 

I don't have a ton of experience with RC2 (more specifically the dotnet runner) but it looks like you got things figured out. I just reviewed this and the results of the appveyor build and it passes the eye test to me. 

Have you tried the resulting package with a .NET 4.5 project/.NET Core project just to make sure the ref's and targets are all good?

joncloud:
Yeah - I just did a re-test using the artifacts from the appveyor build, and the packages seem to be compatible with .NET 4.5 and .NET RC2 projects.

Haacked:
> Yeah - I just did a re-test using the artifacts from the appveyor build, and the packages seem to be compatible with .NET 4.5 and .NET RC2 projects.

Sounds good to me. I'll have to trust you on this one. Anyone else have any concerns or should I go ahead and merge this?

jpsingleton:
The latest CI [package](https://ci.appveyor.com/project/Haacked15676/scientist-net/build/1.0.140/artifacts) still says `alpha5` even though `project.json` was updated to `alpha6`. 
I'm not sure if this is an issue though.

Perhaps it's just an Appveyor setting or maybe the FAKE build script might needs some changes.
The final package from master will need a new name or it will conflict with the [version on NuGet](https://www.nuget.org/packages/scientist/1.0.0-alpha5).

davezych:
`project.json` is going away (this is the best link I could find that actually confirms this, but I do know it's true: https://blogs.msdn.microsoft.com/webdev/2016/05/11/notes-from-the-asp-net-community-standup-may-10-2016/). I don't know if that has anything to do with the package version still showing `alpha5`.

jpsingleton:
This blog post that I linked to in the issue explains the `project.json` removal a bit more:
https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json

I'm not sure what the correct way to package libraries is now. The [docs](https://dotnet.github.io/docs/core-concepts/libraries/libraries-with-cli.html) say: 

> Some details are subject to change as the toolchain evolves.

joncloud:
After a little bit more research I found we're parsing the data from `ReleaseNotes.md`.  I updated it based upon what current data is there, and I suspect the appveyor build should provide the correct version.  Local testing of `build.cmd CreatePackages` provides the correct naming.

davezych:
Oh yes, that thing I did that I forgot about. 

Good job, @joncloud :+1:

Haacked:
So, is this ready to merge? I'm not set up to test it right now.

joncloud:
I think so.

Haacked:
:thumbsup:

build.fsx
```diff
@@ -47,21 +45,8 @@ let Run workingDirectory fileName args =
...
-        if buildServer = BuildServer.AppVeyor
-            then "C:\\Program Files\\Microsoft DNX\\Dnvm\\"
-            else (GetHomeDirectory + "\\.dnx\\bin\\")
-
-    homeDirectory
-
-let GetDnxHome =
-    let homeDirectory = GetHomeDirectory
-    homeDirectory + ("\\.dnx\\runtimes\\dnx-" + runtime + "-win-" + architecture + "." + runtimeVersion + "\\bin\\")
+let dotnetHome = "C:\\Program Files\\dotnet\\"
```
Haacked: I should have reviewed this more carefully. This won't work on machines where `dotnet` doesn't exist. Is there a command to download dotnet.exe? If not, we should put dotnet.exe in the `tools` directory.

joncloud: Let me create a new issue, and start looking into this.

PR #65, opened 2016-06-10
Updated build to download dotnet into tools folder.

### LINKED ISSUE #64: build.cmd fails when dotnet cli is not installed
build.fsx currently requires dotnet.exe to be pre-installed in C:\Program Files\dotnet\.

joncloud:
This should fix #64.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -1,8 +1,10 @@
 #r @"tools/FAKE.Core/tools/FakeLib.dll"
 
 open Fake
+open Fake.FileHelper
 open System
 open System.Collections.Generic
+open System.Net;
 open System.Text
 
 let architecture = getBuildParamOrDefault "architecture" "x86"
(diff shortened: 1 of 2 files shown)


build.fsx
```diff
@@ -45,8 +47,11 @@ let Run workingDirectory fileName args =
 
     ProcessResult.New code messages errors
 
-let dotnetHome = "C:\\Program Files\\dotnet\\"
+let dotnetHome = ".\\tools\dotnet\\"
 let dotnetExe = dotnetHome + "dotnet.exe"
+let dotnetInstall = dotnetHome + "dotnet-install.ps1"
+let dotnetInstallPath = "https://raw.githubusercontent.com/dotnet/cli/rel/1.0.0/scripts/obtain/dotnet-install.ps1"
```
joncloud: I wasn't sure if this should be downloaded like this or added directly to the repository.

Haacked: Downloading it this way keeps the repository smaller. If builds take too long, we can just put it in the directory.

Haacked:
Thanks!

PR #67, opened 2016-06-11
Implemented the ability to keep publishing clean.

joncloud:
Updated experiment classes to take two generic types one for value and clean.
Added overload to Scientist to allow overriding the clean type.
Refactored several parts to reduce code redundancy.


Haacked:
Looks good! Thanks!

PR #68, opened 2016-06-13
Updated result publishing to be fire and forget.

joncloud:
Added ConcurrentSet, which wraps up ConcurrentDictionary calls without needing TValue.
Updated ExperimentInstance to track publishing tasks as fire and forget to avoid publishing from blocking the result's return value.
Added tests to ensure that Science calls are not blocked by result publishing.

### DIFF
--- a/src/Scientist/Internals/ExperimentInstance.cs
+++ b/src/Scientist/Internals/ExperimentInstance.cs
@@ -23,6 +23,7 @@ namespace GitHub.Internals
         internal readonly Func<Task<bool>> RunIf;
         internal readonly IEnumerable<Func<T, T, Task<bool>>> Ignores;
         internal readonly Dictionary<string, dynamic> Contexts;
+        internal readonly ConcurrentSet<Task> PublishingTasks;
         internal readonly Action<Operation, Exception> Thrown;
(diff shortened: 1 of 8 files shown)


src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -88,9 +90,14 @@ public ExperimentInstance(ExperimentSettings<T, TClean> settings)
...
-                // TODO: Make this Fire and forget so we don't have to wait for this
-                // to complete before we return a result
-                await Scientist.ResultPublisher.Publish(result);
+                Task task = Scientist.ResultPublisher.Publish(result);
+                PublishingTasks.TryAdd(task);
+
+                // Disable the warning, because the task is being tracked on the set, 
+                // and immediately removed upon completion.
+#pragma warning disable CS4014
+                task.ContinueWith(_ => PublishingTasks.TryRemove(task));
```
joncloud: I just thought a bit more about this and I think I'll need to address the exception handling a little better here.

joncloud:
#80 has a better design discussion.

Haacked:
Thanks for your contribution! As you know, this PR was superseded by #83. :smile:

PR #73, opened 2016-06-26
Added new tests for experiment without Try

arntj:
Added new test to ensure that experiment returns correctly even without any candidate. Also added tests to verify that the Clean, RunIf and Enabled delegates are not run and no publish is done if there are no candidates.

### DIFF
--- a/test/Scientist.Test/ScientistTests.cs
+++ b/test/Scientist.Test/ScientistTests.cs
@@ -821,7 +821,8 @@ public class TheScientistClass
                 experiment.Use(() => 42);
             });
 
-            Assert.False(TestHelper.Results<int>().Any(m => m.ExperimentName == experimentName));
+            Assert.False(TestHelper.Results<int>().Any(m => m.ExperimentName == experimentName),
+                "When an experiment without a candidate is run, no results should be published.");
         }
     }
 }
--- a/test/Scientist.Test/ScientistTests.cs
+++ b/test/Scientist.Test/ScientistTests.cs
@@ -744,5 +744,84 @@ public class TheScientistClass
 
             mock.Received().Clean(expectedResult);
         }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


test/Scientist.Test/ScientistTests.cs
```diff
@@ -744,5 +744,84 @@ public void KeepingItClean()
...
+        public void NoCandidateMeansNoPublish()
+        {
+            const string experimentName = "noCandidateNoPublish";
+
+            var result = Scientist.Science<int>(experimentName, experiment =>
+            {
+                experiment.Use(() => 42);
+            });
+
+            Assert.False(TestHelper.Results<int>().Any(m => m.ExperimentName == experimentName));
```
joncloud: Consider adding a message to this method call to provide some feedback to the test runner.

arntj: @joncloud Thanks for your suggestion. I've added a message.

test/Scientist.Test/ScientistTests.cs
```diff
@@ -744,5 +744,85 @@ public void KeepingItClean()
 
             mock.Received().Clean(expectedResult);
         }
+
+        [Fact]
+        public void NoCandidate()
```
Haacked: I think the method name should communicate the expected outcome. For example, `NoCandidateStillRunsControl`.

Haacked:
Looks good! One minor nit about a test method name. Also, I merged some changes in master that conflict with this. You'll either need to rebase this against master and force push to your branch OR just merge upstream master in your branch.

arntj:
@Haacked Thanks for your suggestions. I've merged the latest changes and changed the method name as you suggested. :thumbsup:

Haacked:
Thanks!

selfie-0

Issue #80, opened 2016-10-03
Async ResultPublishers could deadlock sync experiments


thematthopkins:
Thanks for the great library!

One issue we ran into is when using an async **ResultPublisher** for synchronous experiments within an MVC app.  

The controller below will deadlock on every request, because of the awaiting on **Task.Delay()** within the synchronous web request.

This article has a more in-depth explanation on why:
http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html 

It looks like **ConfigureAwait** could potentially be used for a fix, or possibly adding a separate **ResultPublisher** method that gets invoked on synchronous calls.  Our current workaround is to use sync implementations of **Publish** without any awaits.

```
using System.Threading.Tasks;
using System.Web.Mvc;
using GitHub;

namespace ScientistDeadlockingResultPublisher.Controllers
{
    public class AsyncResultPublisher : IResultPublisher
    {
        public async Task Publish<T, TClean>(Result<T, TClean> result)
        {
            await Task.Delay(5000);
        }
    }

    public class DeadlockController : Controller
    {
        // GET: Deadlock
        public ActionResult Index()
        {
            Scientist.ResultPublisher = new AsyncResultPublisher();
            var result = Scientist.Science<int>("myExperiment", e =>
            {
                e.Use(() => 2);
                e.Try("candidate", () => 1);
            });

            return Content("done");
        }
    }

```

Haacked:
> It looks like `ConfigureAwait` could potentially be used for a fix

To make sure I understand, you're suggesting that when Scientist calls `IResultPublisher.Publish`, it should append a `ConfigureAwait(false)` call to that?

Could you try it out in your example here by putting `Task.Delay(5000).ConfigureAwait(false);` and make sure that doesn't deadlock?

thematthopkins:
I tested out a couple approaches.  The problem with `ConfigureAwait(false)` is that I had to add it on every point up the call stack.  Meaning:
``Task.Delay(5000).ConfigureAwait(false);```

But also changing `ExperimentInstance` to use
```await Scientist.ResultPublisher.Publish(result).ConfigureAwait(false);  ```

As well as doing the same for the top level invocation from `Scientist`. 

It would also require the implementation of the `ResultPublisher` to include it in all descendant async calls.

It doesn't feel like a great solution, since the `ResultPublisher` may need to invoke some async libraries that likely don't come pre-wired with `ConfigureAwait(false)`.

[...]

PR #83, opened 2016-10-14
Adding FireAndForgetResultPublisher

thematthopkins:
See #80 
Borrows heavily from @joncloud's implementation

### DIFF
--- a/src/Scientist/FireAndForgetResultPublisher.cs
+++ b/src/Scientist/FireAndForgetResultPublisher.cs
@@ -7,7 +7,7 @@ namespace GitHub
     public enum ResultPublisherExceptionBehavior
     {
         Rethrow,
-        Swollow
+        Ignore
     }
 
     public class FireAndForgetResultPublisher : IResultPublisher
--- a/src/Scientist/FireAndForgetResultPublisher.cs
+++ b/src/Scientist/FireAndForgetResultPublisher.cs
@@ -7,13 +7,13 @@ namespace GitHub
     public class FireAndForgetResultPublisher : IResultPublisher
     {
 		private readonly IResultPublisher _publisher;
-		private readonly Action<Exception> _onThrown;
+		private readonly Action<Exception> _onPublisherException;
 		private readonly ConcurrentSet<Task> _publishingTasks = new ConcurrentSet<Task>();
 
-		public FireAndForgetResultPublisher(IResultPublisher publisher, Action<Exception> onThrown = null)
+		public FireAndForgetResultPublisher(IResultPublisher publisher, Action<Exception> onPublisherException = null)
 		{
 			_publisher = publisher;
-			_onThrown = onThrown;
+			_onPublisherException = onPublisherException;
 		}
 
 		public Task Publish<T, TClean>(Result<T, TClean> result)
@@ -30,8 +30,8 @@ namespace GitHub
 				}
 				catch(Exception ex)
 				{
-					if (_onThrown != null)
-						_onThrown(ex);
+					if (_onPublisherException != null)
+						_onPublisherException(ex);
 					else
 						throw;
 				}
--- a/src/Scientist/FireAndForgetResultPublisher.cs
+++ b/src/Scientist/FireAndForgetResultPublisher.cs
@@ -10,6 +10,10 @@ namespace GitHub
         private readonly Action<Exception> _onPublisherException;
         private readonly ConcurrentSet<Task> _publishingTasks = new ConcurrentSet<Task>();
 
+        public FireAndForgetResultPublisher(IResultPublisher publisher) : this(publisher, e => { })
+        {
+        }
+
         public FireAndForgetResultPublisher(IResultPublisher publisher, Action<Exception> onPublisherException)
         {
             _publisher = publisher;
--- a/src/Scientist/FireAndForgetResultPublisher.cs
+++ b/src/Scientist/FireAndForgetResultPublisher.cs
@@ -1,0 +1,51 @@
+﻿using GitHub.Internals;
+using System;
+using System.Threading.Tasks;
+
+namespace GitHub
+{
+    public class FireAndForgetResultPublisher : IResultPublisher
+    {
+		private readonly IResultPublisher _publisher;
+		private readonly Action<Exception> _onThrown;
+		private readonly ConcurrentSet<Task> _publishingTasks = new ConcurrentSet<Task>();
+
+		public FireAndForgetResultPublisher(IResultPublisher publisher, Action<Exception> onThrown = null)
+		{
+			_publisher = publisher;
+			_onThrown = onThrown;
+		}
+
+		public Task Publish<T, TClean>(Result<T, TClean> result)
+		{
+
+			// Disable the warning, because the task is being tracked on the set, 
+			// and immediately removed upon completion.
+#pragma warning disable CS4014
+			var subTask = Task.Run(async () =>
+			{
+				try
+				{
+					await _publisher.Publish(result);
+				}
+				catch(Exception ex)
+				{
+					if (_onThrown != null)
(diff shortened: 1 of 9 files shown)


thematthopkins:
I'm not sure on what would be the best behavior for exceptions would be.  Implicitly assuming that the exception should be re-thrown on omission of `onPublisherException` doesn't seem right.  I added a `ResultPublisherExceptionBehavior` enum to make behavior explicit.

src/Scientist/FireAndForgetResultPublisher.cs
```diff
@@ -0,0 +1,58 @@
+﻿using GitHub.Internals;
+using System;
+using System.Threading.Tasks;
+
+namespace GitHub
+{
+    public enum ResultPublisherExceptionBehavior
+    {
+        Rethrow,
+        Swollow
```
Haacked: Swallow is mispelled. 😛 Also, I'd prefer it be `Ignore`.

src/Scientist/Internals/ConcurrentSet.cs
```diff
@@ -0,0 +1,34 @@
+﻿using System.Collections.Concurrent;
+using System.Collections.Generic;
+
+namespace GitHub.Internals
+{
+	internal class ConcurrentSet<T>
+	{
+		static readonly byte One = 1;
```
Haacked: Minor nit, but this entire implementation looks indented too much. We indent 4 spaces.

thematthopkins: My mistake.  Tabs snuck in there.  Thanks!  Fixed

src/Scientist/FireAndForgetResultPublisher.cs
thematthopkins: Thanks! Changed to Ignore + added as a default

src/Scientist/FireAndForgetResultPublisher.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                try
+                {
+                    await _publisher.Publish(result);
+                }
+                catch(Exception ex)
+                {
+                    _onPublisherException?.Invoke(ex);
+                    if(_exceptionBehavior == ResultPublisherExceptionBehavior.Rethrow)
+                        throw;
```
thematthopkins: Our use-case these unhandled exceptions bubble up to IIS and end up in the Windows event log, so they aren't totally missed.  Unhandled here is the simplest thing to do for us, though that may not be a typical use-case and it does create an untestable branch.

src/Scientist/FireAndForgetResultPublisher.cs
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    public class FireAndForgetResultPublisher : IResultPublisher
+    {
+        private readonly IResultPublisher _publisher;
+        private readonly Action<Exception> _onPublisherException;
+        private readonly ConcurrentSet<Task> _publishingTasks = new ConcurrentSet<Task>();
+        private readonly ResultPublisherExceptionBehavior _exceptionBehavior;
+
+        public FireAndForgetResultPublisher(IResultPublisher publisher, ResultPublisherExceptionBehavior exceptionBehavior = ResultPublisherExceptionBehavior.Ignore, Action<Exception> onPublisherException = null)
```
Haacked: Optional arguments in public APIs can cause versioning issues down the road: http://haacked.com/archive/2010/08/10/versioning-issues-with-optional-arguments.aspx/

We should use constructor overloads instead.

src/Scientist/FireAndForgetResultPublisher.cs
```diff
@@ -0,0 +1,58 @@
...
+#pragma warning disable CS4014
+            var subTask = Task.Run(async () =>
+            {
+                try
+                {
+                    await _publisher.Publish(result);
+                }
+                catch(Exception ex)
+                {
+                    _onPublisherException?.Invoke(ex);
```
Haacked: We use `NullGuard` on this project so `_onPublisherException` can't be null here. If we want to give consumers the ability to pass in null for this, we need to us the `[AllowNull]` attribute on the constructor argument.

src/Scientist/FireAndForgetResultPublisher.cs
Haacked: Right now, our code requires someone pass in an `_onPublisherException` handler. If we allowed that to be null, then maybe I'd understand also having this option. As it stands, if you want this exception logged, do it in the `_onPublisherException` callback.

src/Scientist/FireAndForgetResultPublisher.cs
thematthopkins: Thanks!  Great point.

src/Scientist/FireAndForgetResultPublisher.cs
thematthopkins: That works for me.  Removed the `ResultPublisherExceptionBehavior` option and made the onPublisherException mandatory to emphasis it.

src/Scientist/FireAndForgetResultPublisher.cs
```diff
@@ -0,0 +1,48 @@
...
+
+namespace GitHub
+{
+    public class FireAndForgetResultPublisher : IResultPublisher
+    {
+        private readonly IResultPublisher _publisher;
+        private readonly Action<Exception> _onPublisherException;
+        private readonly ConcurrentSet<Task> _publishingTasks = new ConcurrentSet<Task>();
+
+        public FireAndForgetResultPublisher(IResultPublisher publisher, Action<Exception> onPublisherException)
```
thematthopkins: Works for me

test/Scientist.Test/FireAndForgetResultPublisherTests.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        // Ensure that the mock was called before the when published task has completed.
+        mock.Received(count).Control();
+        mock.Received(count).Candidate();
+
+        Assert.False(whenPublished.IsCompleted, "When Published Task completed early.");
+
+        await whenPublished;
+
+        Assert.True(fireAndForgetPublisher.WhenPublished().IsCompleted, "When Published Task isn't complete.");
```
Haacked: Shouldn't we just assert `whenPublished.IsCompleted` here? As in:

`Assert.True(whenPublished.IsCompleted, "When Published Task isn't complete.");`

test/Scientist.Test/FireAndForgetResultPublisherTests.cs
```diff
@@ -0,0 +1,132 @@
...
+    [Fact]
+    public async Task PublishesAsynchronously()
+    {
+        const int expectedResult = 42;
+
+        // Create a new publisher that will delay all
+        // publishing to account for this test.
+        var innerPublisher = Substitute.For<IResultPublisher>();
+        innerPublisher.Publish(Arg.Any<Result<int, int>>())
+            .Returns(call => Task.Delay(100));
```
Haacked: I'm a little antsy about using delays for tests because they take a little bit of determinism away from the test. For example, some weird timing issue could in theory cause this test to fail.

Instead, I generally like to use something like a `ManualResetEvent` here to tightly control order of events. Then we could call `Signal` on the event before we `await whenPublished`. Thoughts?

One of the reasons I note this is this test runs significantly slower than all the others because of this artificial delay.

BTW, the same approach probably applies to `HandlesDelayedExceptionsThrownByInnerPublisher`

test/Scientist.Test/FireAndForgetResultPublisherTests.cs
thematthopkins: fixed

Haacked:
Nice work! It was a team effort! :sparkles:

Issue #59, opened 2016-05-04
SingleContextIncludedWithPublish Unit Test passes inconsistently


joncloud:
Running build.cmd or tests through Visual Studio seems to occasionally fail due to an exception.

Test run on 43e27ee6b42fb6c34ec174df1b46a1b7137a13a8

```
xUnit.net DNX Runner (32-bit DNX 4.5.1)
  Discovering: Scientist.Test
  Discovered:  Scientist.Test
  Starting:    Scientist.Test
    TheScientistClass+TheScienceMethod.SingleContextIncludedWithPublish [FAIL]
      System.AggregateException : One or more errors occurred.
      ---- System.Exception : Exception of type 'System.Exception' was thrown.
      Stack Trace:
           at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
           at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
           at System.Threading.Tasks.Task`1.get_Result()
        C:\Users\jdber_000\Documents\GitHub\scientist.net\src\Scientist\Scientist.cs(35,0): at GitHub.Scientist.Science[T](String name, Action`1 experiment)
        C:\Users\jdber_000\Documents\GitHub\scientist.net\test\Scientist.Test\ScientistTests.cs(539,0): at TheScientistClass.TheScienceMethod.SingleContextIncludedWithPublish()
        ----- Inner Stack Trace -----
        C:\Users\jdber_000\Documents\GitHub\scientist.net\src\Scientist\Internals\Experiment.cs(13,0): at GitHub.Internals.Experiment`1.<>c.<.cctor>b__35_1(Operation operation, Exception exception)
        C:\Users\jdber_000\Documents\GitHub\scientist.net\src\Scientist\Internals\ExperimentInstance.cs(87,0): at GitHub.Internals.ExperimentInstance`1.<Run>d__12.MoveNext()
  Finished:    Scientist.Test
=== TEST EXECUTION SUMMARY ===
   Scientist.Test  Total: 34, Errors: 0, Failed: 1, Skipped: 0, Time: 0.554s
```

Haacked:
Is this still ongoing?

joncloud:
I'll review the code in its current state to see if its applicable, and run some more tests to make sure.

joncloud:
[...]

PR #84, opened 2016-12-08
Update tests to run in serial for #59.

joncloud:
Tests rely on static properties of Scientist to consistently run in completion.  This ensures that each test will perform all necessary changes to the Scientist static properties, and complete without another test updating the static properties into another state.


Haacked:
👍 Thanks!

Issue #85, opened 2016-12-08
Build script fails to run tests with dotnet cli preview5 (latest)


joncloud:
It looks like the latest version of dotnet cli requires a different set of parameters in order to run the unit tests.

A couple of ideas:
* Specify `-Version 1.0.0-preview2-003121` when downloading dotnet cli
* Continue developing #71

Another thing that should probably be updated is that if the dotnet test command fails it should fail the build instead of allowing it to pass without checking tests.

Sample output from [appveyor]:

[appveyor]: https://ci.appveyor.com/project/Haacked/scientist-net/build/1.0.197#L143

```bash
.\tools\dotnet\dotnet.exe test .\test\Scientist.Test\
Couldn't find a project to run test from. Ensure a project exists in C:\projects\scientist-net.
Or pass the path to the project
Finished Target: RunTests
```

Haacked:
> Specify -Version 1.0.0-preview2-003121 when downloading dotnet cli

I like that idea as that should work immediately while we continue to work on #71 right?

> Another thing that should probably be updated is that if the dotnet test command fails it should fail the build instead of allowing it to pass without checking tests.

Agreed.

joncloud:
Yup - we can definitely work on these in conjunction.

joncloud:
This was fixed on #86.  It looks like I didn't use the keyword for closing the issue in the commit message.

PR #86, opened 2016-12-09
Fix build script to catch unit test failures for #85.

joncloud:
Update dotnet cli version to match preview2 for compatibility with project.
Correct typo in folder name.
Add check to make sure dotnet test returns a 0 exit code so that if the test fails at all the build process fails altogether.


joncloud:
I wasn't entirely sure about the syntax/behavior in for checking the exit code, but this seemed to work locally.

Haacked:
Good enough for me. :smile: Thanks!

PR #88, opened 2017-01-19
Added csproj files

### LINKED ISSUE #71: Switch back to .csproj from project.json
So given [this news](https://blogs.msdn.microsoft.com/dotnet/2016/05/23/changes-to-project-json/) we should probably switch back.

andrewhart098:
Using VS 2017, I open Scientist.net and went through the conversion process from project.json to csproj.

When I built the project, it initially complained about not being able to find the file xunit.runner.json - so I removed that missing file from the project.

The tests pass using the command `dotnet test Scientist.Test.csproj`

Fixes #71


joncloud:
It looks like the build is still running against the original `project.json` and `*.xproj`.  Are there any pitfalls to having both project definitions (new and old) concurrently?

joncloud:
It looks like that change didn't remove the `project.json` files from the repository.  FYI I created a [prototype](https://github.com/joncloud/Scientist.net/commit/4c245745a202d836f6eb8812b9fc5a25e03f0fd0) for converting to csproj.  It looks like there have been more releases of VS/dotnet which may alter the structure.  Feel free to take a look over that as a base.

andrewhart098:
Hey @joncloud, thanks for the information.  I noticed the changes to `build.fsx` in your prototype project, should that be changed with this PR or in a separate PR?

I went through an actually removed the project.json files from the repo. I can still build and the tests pass on my machine.  

The CI seems to still want to run everything like the project.json is still there: From appveyor log:
`System.InvalidOperationException: .\test\Scientist.Test\project.json does not exist.`

joncloud:
Yeah - In order to get the build process working correctly I think the change to `build.fsx` will be necessary.  You'll notice that it uses the .NET CLI to restore/build/package the project, and there are references to `project.json`.  I think that's what is causing the failure in the AppVeyor integration.

Haacked:
Any progress on this? Anything I can do to help it along?

joncloud:
@Haacked I re-ran the conversion with the latest release of VS2017 on #95.

PR #94, opened 2017-03-16
Documentation for async parallel calls

jon-adams:
Documentation for async parallel calls—RE: [pull request #66](https://github.com/github/Scientist.net/pull/66) that fixed [issue #62](https://github.com/github/Scientist.net/issues/62)

Also threw in a minor clarification for the new feature of `FireAndForgetResultPublisher` that is in the master branch but is not released to NuGet yet, which was causing some confusion.


Haacked:
Thanks for the documentation update!

PR #95, opened 2017-03-17
Initial conversion to csproj for #71.

joncloud:
Convert src and test projects using VS2017.
Update VS solution using VS2017.
Update build script to reference csproj/nuspec instead of directory/project.json.


Haacked:
I took some other work you and others did and got it working. Please check my work. I think this PR is no longer needed.

joncloud:
@Haacked I found a couple of minor issues that stuck out after running some tests with VS2017.  I opened up a couple of PRs for resolving them.

PR #98, opened 2017-07-12
Included information about the operation type that threw the exception.

Silvenga:
When an error occurs within the publisher (and other locations), to a new developer, the stack trace may be confusing, e.g.

```
System.AggregateException: One or more errors occurred. ---> System.Collections.Generic.KeyNotFoundException: The given key was not present in the dictionary.
   at GitHub.Internals.Experiment`2.<>c.<.cctor>b__39_1(Operation operation, Exception exception)
   at GitHub.Internals.ExperimentInstance`2.<Run>d__15.MoveNext()
```

This pull includes additional information when exceptions are thrown using default configurations.

### DIFF
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -10,7 +10,7 @@ namespace GitHub.Internals
 
         private static readonly Func<Task<bool>> _alwaysRun = () => Task.FromResult(true);
         private static readonly Action<Operation, Exception> _alwaysThrow
-            = (operation, exception) => throw new OperationException(operation, exception);
+            = (operation, exception) => { throw new OperationException(operation, exception); };
 
         private string _name;
         private int _concurrentTasks;
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -10,25 +10,7 @@ namespace GitHub.Internals
 
         private static readonly Func<Task<bool>> _alwaysRun = () => Task.FromResult(true);
         private static readonly Action<Operation, Exception> _alwaysThrow
-            = (operation, exception) =>
-            {
-                string exceptionMessage;
-                switch (operation)
-                {
-                    case Operation.Clean:
-                    case Operation.Compare:
-                    case Operation.Enabled:
-                    case Operation.Ignore:
-                    case Operation.Publish:
-                    case Operation.RunIf:
-                        exceptionMessage = $"An exception occurred within the experiment during the '{operation.ToString()}' operation.";
-                        break;
-                    default:
-                        exceptionMessage = "An exception occurred within the experiment, the operation was unknown.";
-                        break;
-                }
-                throw new Exception(exceptionMessage, exception);
-            };
+            = (operation, exception) => throw new OperationException(operation, exception);
 
         private string _name;
         private int _concurrentTasks;
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -10,7 +10,25 @@ namespace GitHub.Internals
 
         private static readonly Func<Task<bool>> _alwaysRun = () => Task.FromResult(true);
         private static readonly Action<Operation, Exception> _alwaysThrow
-            = (operation, exception) => { throw exception; };
+            = (operation, exception) =>
+            {
+                string exceptionMessage;
+                switch (operation)
+                {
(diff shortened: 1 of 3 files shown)


src/Scientist/Internals/Experiment.cs
```diff
@@ -10,7 +10,25 @@ internal class Experiment<T, TClean> : IExperiment<T, TClean>, IExperimentAsync<
 
         private static readonly Func<Task<bool>> _alwaysRun = () => Task.FromResult(true);
         private static readonly Action<Operation, Exception> _alwaysThrow
-            = (operation, exception) => { throw exception; };
+            = (operation, exception) =>
+            {
+                string exceptionMessage;
+                switch (operation)
```
davezych: Is there ever a case where we _won't_ know the operation in question? Seems to me like this switch isn't needed, and we can always throw with the `exceptionMessage = $"An exception occurred within the experiment during the '{operation.ToString()}' operation.";` message.

joncloud: Agreed - I think if we add another `Operation` to the list then we should be using it explicitly, and it should be known.

src/Scientist/Internals/Experiment.cs
```diff
@@ -10,7 +10,25 @@ internal class Experiment<T, TClean> : IExperiment<T, TClean>, IExperimentAsync<
...
+                    case Operation.Ignore:
+                    case Operation.Publish:
+                    case Operation.RunIf:
+                        exceptionMessage = $"An exception occurred within the experiment during the '{operation.ToString()}' operation.";
+                        break;
+                    default:
+                        exceptionMessage = "An exception occurred within the experiment, the operation was unknown.";
+                        break;
+                }
+                throw new Exception(exceptionMessage, exception);
```
joncloud: Does it make sense to create a new `OperationException` type for this situation?  We could pass the specific `Operation`, `innerException` and then internally handle the rest.  This would allow the caller if necessary to catch and handle the specific `Operation` instead of having to parse the message.

Silvenga: I was just thinking about doing that, I'll implement that.

src/Scientist/Internals/Experiment.cs
Silvenga: I see your point, didn't think about that. I've moved the check to a simple `IsDefined`, although, yes, we should only be using defined enums within the code base.

src/Scientist/OperationException.cs
```diff
@@ -0,0 +1,31 @@
...
+    public class OperationException : Exception
+    {
+        /// <summary>
+        /// The operation in which this exception occurred within.
+        /// </summary>
+        public Operation Operation { get; }
+
+        public override string Message { get; }
+
+        internal OperationException(Operation operation, Exception inner) : base(null, inner)
```
joncloud: Should we use the base class constructor to assign the exception's message here, i.e., `base(GetExceptionMessage(operation), inner)`?

joncloud: I think that might prevent the need for overriding the `Message` property on this class specifically.

Silvenga:
Not sure where this issue is with AppVeyor - that should be completely valid C#. 

VS17 can build the project - although Resharper is freaking out about other things. I can't actually edit the project without auto-upgrading the xproj to csproj.

src/Scientist/OperationException.cs
Silvenga: Of course, I was more wondering the proper way of creating exceptions under .NetStandard.

src/Scientist/Internals/Experiment.cs
```diff
@@ -10,7 +10,7 @@ internal class Experiment<T, TClean> : IExperiment<T, TClean>, IExperimentAsync<
 
         private static readonly Func<Task<bool>> _alwaysRun = () => Task.FromResult(true);
         private static readonly Action<Operation, Exception> _alwaysThrow
-            = (operation, exception) => { throw exception; };
+            = (operation, exception) => throw new OperationException(operation, exception);
```
joncloud: It looks like AppVeyor is barfing here.  I think throw expressions were added in C# 7.0, which isn't yet available in this project.  Adding the brackets back here should resolve that issue.

joncloud:
Looks great!  Do you think we can write some unit tests to verify the exception that gets thrown from the default implementation?

Haacked:
Nice work! Sorry it took me so long to get to this.

PR #100, opened 2017-08-10
Improves performance while timing observerations.

joncloud:
Reduces operations by calculating duration in line.
Removes an allocation per observation.

### DIFF
--- a/src/Scientist/Observation.cs
+++ b/src/Scientist/Observation.cs
@@ -110,13 +110,14 @@ namespace GitHub
             return observation;
         }
 
+        private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
         /// <summary>
         /// Runs the experiment.
         /// </summary>
         internal async Task Run(Func<Task<T>> block)
         {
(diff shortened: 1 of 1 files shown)


davezych:
So I have no problem with this, looks benign enough.

However my only question is why - sure, reducing an allocation is good, but I doubt it really had much of an affect. Did you run into a situation where this was causing harm? Just curious, approved on my end either way.

joncloud:
I didn't run into a situation where it was causing harm.  I ended up writing some code in another application which needed to do timing.  I reviewed how ASP.NET Core was doing it, and implemented it very similar to this.  I had noticed that we were creating the `Stopwatch` instance and figured it would be worthwhile to update too.

I also ran some [performance testing](https://gist.github.com/joncloud/f217d8946cd94ff0ccb4e33f790247b9) using BenchmarkDotNet.

davezych:
Neat 👍 

Looks good, :shipit: (cc @Haacked )

Haacked:
Reducing allocations is always a good thing. :smile: I agree that there's probably no specific problem here, but it doesn't hurt to follow patterns the ASP.NET team are doing.

PR #101, opened 2017-08-11
fixes #99 - only throw inner exception for AggregateExceptions

### LINKED ISSUE #99: Option to re-throw original exception instead of inner exception?
[Observation.cs#L126](https://github.com/github/Scientist.net/blob/master/src/Scientist/Observation.cs#L126)

We're losing a lot of information when this re-throws the inner exception rather than the original exception.

Open to always using the original exception or adding an option somewhere to use the original exception instead?

FUR10N:
Tried to add a test, but since the tests were all synchronous, everything got wrapped in AggregateException no matter what I did.


joncloud:
Is it possible for us to add an asynchronous test to exercise the new code?

FUR10N:
Yep, got one working.

Haacked:
Nice work! I'll fix this up and get it merged.

PR #105, opened 2017-10-05
Fixes build warnings.

joncloud:
Minor refactor to use generic on method instead of implicitly defining a return value, which would never get executed.


Haacked:
:thumbs up:

PR #111, opened 2018-07-02
Added AllocatedBytes metrics to Observations

Lexcess:
Hi, 

I quite like the idea of Scientist.Net, as I often don't want to set up a full Benchmark.Net run. However I found the lack of memory allocation information a bit of showstopper.   

This Pull Request adds a new `AllocatedBytes `field to Observation. To enable this to work cross runtime I've cribbed a lot of the hackery from Benchmark.Net's implementation (Garbage Collection tooling is wildly inconsistent - see the last paragraph). Included are a test for the basic functionality and one that asserts that Scientist's own allocations don't appear in the results. Note that I did not include the Generation collection counts because it was trivially easy to get miscounts with the way Scientist.Net works versus Benchmark.Net's dedicated runs.

I also upped the versions of the Nuget package (for my testing), the Core SDK (for the exe/unit tests) so they used the latest GC Tooling. You can revert these changes and the code will still work.

In terms of impact of adding this behaviour. For .Net Core it is almost imperceptible, for .Net Framework the only way to achieve this functionality is with `GC.Collect`s which is a little more invasive. I would say that the benefit of seeing allocations outweighs the downsides, especially since allocations are more consistent than times for Scientist's kind of micro-benchmarking. 

Lastly I've [proposed an addition to vNext of .Net Standard](https://github.com/dotnet/standard/issues/771) that would do away with all the hackery for both this, Benchmark.Net and anyone else who cares about allocations at runtime.  

### Commit notes

```
Added target for .Net Standard 2.0 to detect better GC tooling
Added test for new AllocatedBytes feature
Added test to account for Scientist's own allocations in AllocatedBytes
Upgraded test and console app to netcoreapp 2.1 to use new GC Tooling
Some portions of GCHelper were derived from Benchmark.Net and are used under license
```

### DIFF
--- a/src/Scientist/Internals/GCHelper.cs
+++ b/src/Scientist/Internals/GCHelper.cs
@@ -1,0 +1,132 @@
+﻿namespace GitHub.Internals
+{
+    using System;
+    using System.Reflection;
+    using System.Runtime.InteropServices;
+
+    internal static class GCHelper
+    {
+        private const string DebugConfigurationName = "DEBUG";
+        internal const string ReleaseConfigurationName = "RELEASE";
+        internal const string Unknown = "?";
+
+        public static bool IsMono { get; } = Type.GetType("Mono.Runtime") != null;
+
+        public static bool IsFullFramework =>
+#if (NET451)
+            true;
+#else
+            System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Framework", StringComparison.OrdinalIgnoreCase);
+#endif
+
+        public static bool IsNetCore =>
+#if (NET451)
+            false;
+#else
+            System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase);
+#endif
+
+        private static long AllocationQuantum { get; } = CalculateAllocationQuantumSize();
+
+        private static readonly Func<long> GetAllocatedBytesForCurrentThreadDelegate = GetAllocatedBytesForCurrentThread();
+
+        /// <summary>
+        /// returns total allocated bytes (not per operation)
+        /// </summary>
+        /// <param name="excludeAllocationQuantumSideEffects">Allocation quantum can affecting some of our nano-benchmarks in non-deterministic way.
+        /// when this parameter is set to true and the number of all allocated bytes is less or equal AQ, we ignore AQ and put 0 to the results</param>
+        /// <returns></returns>
+        public static long GetAllocatedBytes()
+        {
+            var allocation = GetAllocatedBytesByRuntime();
+
+            if (IsNetCore)
+            {
+                return allocation;
+            }
+
+            return allocation <= AllocationQuantum ? 0L : allocation;
+        }
+
+        private static long GetAllocatedBytesByRuntime()
+        {
+            // Monitoring is not available in Mono, see http://stackoverflow.com/questions/40234948/how-to-get-the-number-of-allocated-bytes-
+            if (IsMono)
+            {
+                return 0;
+            }
+            //GC.Collect();
+
+#if (NET451)
+            // "This instance Int64 property returns the number of bytes that have been allocated by a specific 
+            // AppDomain. The number is accurate as of the last garbage collection." - CLR via C#
+            // so we enforce GC.Collect here just to make sure we get accurate results
+            return AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize;
+#elif (NETSTANDARD2_0)
+            // it can be a .NET app consuming our .NET Standard package
+            if (IsFullFramework) 
+            {
+                GC.Collect();
+                return AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize;
+            }
+            // https://apisof.net/catalog/System.GC.GetAllocatedBytesForCurrentThread() is not part of the .NET Standard, so we use reflection to call it..
+            return GetAllocatedBytesForCurrentThreadDelegate.Invoke();
+#elif (NETCOREAPP2_1)
+            // but CoreRT does not support the reflection yet, so only because of that we have to target .NET Core 2.1
+            // to be able to call this method without reflection and get MemoryDiagnoser support for CoreRT ;)
+            return GC.GetAllocatedBytesForCurrentThread();
+#else
+            return 0;
+#endif
+        }
+
+        private static Func<long> GetAllocatedBytesForCurrentThread()
+        {
+            // for some versions of .NET Core this method is internal, 
+            // for some public and for others public and exposed ;)
+#if (NETSTANDARD2_0 || NETCOREAPP2_1)
+            var method = typeof(GC)
+                            .GetTypeInfo()
+                            .GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.Public | BindingFlags.Static)
+                         ?? typeof(GC)
+                            .GetTypeInfo()
+                            .GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.NonPublic | BindingFlags.Static);
+            return () => (long)method.Invoke(null, null);
+#else
+            return () => 0L;
+#endif
+        }
+
+        /// <summary>
+        /// code copied from https://github.com/rsdn/CodeJam/blob/71a6542b6e5c52ea8dd92c601adad11e62796a98/PerfTests/src/%5BL4_Configuration%5D/Metrics/%5BMetricValuesProvider%5D/GcMetricValuesProvider.cs#L63-L89
+        /// </summary>
+        /// <returns></returns>
+        private static long CalculateAllocationQuantumSize()
+        {
+            long result;
+            int retry = 0;
+            do
+            {
+                if (++retry > 10)
+                {
+                    // 8kb https://github.com/dotnet/coreclr/blob/master/Documentation/botr/garbage-collection.md
+                    result = 8192;
+                    break;
+                }
+
+                GC.Collect();
+                GC.WaitForPendingFinalizers();
+                GC.Collect();
+
+                result = GC.GetTotalMemory(false);
+                var tmp = new object();
+                result = GC.GetTotalMemory(false) - result;
+                GC.KeepAlive(tmp);
+            } while (result <= 0);
+
+            return result;
+        }
+    }
+}
--- a/src/Scientist/Observation.cs
+++ b/src/Scientist/Observation.cs
@@ -1,4 +1,5 @@
-﻿using System;
+﻿using GitHub.Internals;
+using System;
 using System.Diagnostics;
 using System.Threading.Tasks;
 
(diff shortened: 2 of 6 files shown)


Lexcess:
Can anyone take a look?

Haacked:
Thanks @Lexcess! I've been on vacation the last two weeks, hence the delay. Are you currently using this in production anywhere? @joncloud, @martincostello, @ryangribble do any of you have time to look at this too?

I'll try and find time to review it this week, but would welcome other eyes on it.

Haacked:
> For .Net Core it is almost imperceptible, for .Net Framework the only way to achieve this functionality is with GC.Collects which is a little more invasive.

This is a concern. Perhaps we should add a flag to enable this. That way it's off by default and for those who want to track allocated bytes, they can turn it on easily.

joncloud:
Sure thing - I'll try and provide some feedback where I can.

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+            return 0;
+#endif
+        }
+
+        private static Func<long> GetAllocatedBytesForCurrentThread()
+        {
+            // for some versions of .NET Core this method is internal, 
+            // for some public and for others public and exposed ;)
+#if (NETSTANDARD2_0 || NETCOREAPP2_1)
+            var method = typeof(GC)
```
joncloud: Can we leverage `MethodInfo.CreateDelegate(typeof(Func<long>))` to create a delegate from the `MethodInfo`? I'm unsure if there are any limitations based on .NET Framework vs .NET Core.

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+            //GC.Collect();
+
+#if (NET451)
+            // "This instance Int64 property returns the number of bytes that have been allocated by a specific 
+            // AppDomain. The number is accurate as of the last garbage collection." - CLR via C#
+            // so we enforce GC.Collect here just to make sure we get accurate results
+            return AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize;
+#elif (NETSTANDARD2_0)
+            // it can be a .NET app consuming our .NET Standard package
+            if (IsFullFramework)
```
joncloud: What do you think about taking this check outside of this method, and use it during the creation of `GetAllocatedBytesForCurrentThreadDelegate`? We'd be able to save the check, and then always just run
```csharp
GC.Collect();
return AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize;
```

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+#endif
+
+        public static bool IsNetCore =>
+#if (NET451)
+            false;
+#else
+            System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Core", StringComparison.OrdinalIgnoreCase);
+#endif
+
+        private static long AllocationQuantum { get; } = CalculateAllocationQuantumSize();
```
joncloud: Do we need to calculate the `AllocationQuantum` for .NET Core? It looks like we only use this for `!IsNetCore`.

joncloud:
This seems fairly solid, however I am definitely not a master of memory. I think @Haacked has a good idea by opting into this behavior. Looking at how you have this implemented I think that could be pretty easy to transition to configuration.

Lexcess:
@Haacked @joncloud I would be fine with it being opt in, I think you'd need to give some steer on how you'd want that to work.

Just as a counterpoint though, I'm hopeful that .Net FX will incorporate per thread allocation data at some point (the Mono team just made a soft commitment to do so on their thread tracking this). So if the issue is the GC.Collect behaviour then perhaps it would be better to add the flag around that rather than the feature itself? In terms of benchmarking memory allocation data is far more reliable than timings so it might be a shame to have it only opt in when the issue is more around one platform (and a platform where perf is less of a focus anyway). 

Again though if you think opt in is the model if you could describe how you'd like it to be enabled then I'll make the change. Could we leave the field on Observation in either case? It would make doing publishing a lot easier (otherwise I guess we'd need it to be subclassed or perhaps a dictionary for optional data)

src/Scientist/Internals/GCHelper.cs
Lexcess: Until recently you did need this for .Net core but yes we can optimised this out here.

src/Scientist/Internals/GCHelper.cs
Lexcess: So do it purely on #IfDef? The reason we do a runtime framework check is because the if you used the code within a dotnet Standard library you wouldn't know at compile time your runtime (FX, Core or Mono). Apologies if I've missed your point here I'm not convinced that GitHub is pointing me to the right line.

src/Scientist/Internals/GCHelper.cs
Lexcess: The issue here is that the .Net Core team soft launched the new API feature as internal before finally making it public sometime later. This code is looking for the method in either form (hence the public/nonpublic binding flags). You could do it the other way, but you'd drop support prior to Net Core 2.1.

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+        internal const string ReleaseConfigurationName = "RELEASE";
+        internal const string Unknown = "?";
+
+        public static bool IsMono { get; } = Type.GetType("Mono.Runtime") != null;
+
+        public static bool IsFullFramework =>
+#if (NET451)
+            true;
+#else
+            System.Runtime.InteropServices.RuntimeInformation.FrameworkDescription.StartsWith(".NET Framework", StringComparison.OrdinalIgnoreCase);
```
martincostello: Just checking, these aren't strings that can vary if language packs are installed on the local machine?

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        private static long GetAllocatedBytesByRuntime()
+        {
+            // Monitoring is not available in Mono, see http://stackoverflow.com/questions/40234948/how-to-get-the-number-of-allocated-bytes-
+            if (IsMono)
+            {
+                return 0;
+            }
+            //GC.Collect();
```
martincostello: Leftover from a refactor?

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+
+#if (NET451)
+            // "This instance Int64 property returns the number of bytes that have been allocated by a specific
```
martincostello: Link to the source of this quote?

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+#elif (NETSTANDARD2_0)
+            // it can be a .NET app consuming our .NET Standard package
+            if (IsFullFramework) 
+            {
+                GC.Collect();
+                return AppDomain.CurrentDomain.MonitoringTotalAllocatedMemorySize;
+            }
+            // https://apisof.net/catalog/System.GC.GetAllocatedBytesForCurrentThread() is not part of the .NET Standard, so we use reflection to call it..
+            return GetAllocatedBytesForCurrentThreadDelegate.Invoke();
+#elif (NETCOREAPP2_1)
```
martincostello: Is this for future use? The library doesn't target `netcoreapp2.1`, so this would never be true at compile-time. That said, it's probably best that the library never targets the app TFMs, so it'd be `NETSTANDARD2_1` (or whatever the next standard version becomes), so until that arrives this `#elif` is redundant.

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+#else
+            return 0;
+#endif
+        }
+
+        private static Func<long> GetAllocatedBytesForCurrentThread()
+        {
+            // for some versions of .NET Core this method is internal, 
+            // for some public and for others public and exposed ;)
+#if (NETSTANDARD2_0 || NETCOREAPP2_1)
```
martincostello: As-per the above comment regarding TFMs.

src/Scientist/Internals/GCHelper.cs
Lexcess: Yep Good catch.

src/Scientist/Internals/GCHelper.cs
Lexcess: Wasn't linked as mentioned I reworked the code from Benchmark.Net for the GC portion of this PR. You could perhaps look at their commits to try and find out.

src/Scientist/Internals/GCHelper.cs
Lexcess: So I used this in development stage so technically it could be removed. I'd suggest keeping it unless it causes a problem though. Might also be useful in the future if people want ILMerge/Native compilation support.

src/Scientist/Internals/GCHelper.cs
Lexcess: Ditto on these

src/Scientist/Internals/GCHelper.cs
Lexcess: Also Hey Martin, I thought that looked like your name in the commits but thought it was a coincidence.

src/Scientist/Internals/GCHelper.cs
joncloud: Yeah - this may have been a little misleading. What i'm suggesting is moving this logic out to where we construct the `GetAllocatedBytesForCurrentThreadDelegate` delegate. I'll add some comments in that area to help facilitate.

Haacked:
> So if the issue is the GC.Collect behaviour then perhaps it would be better to add the flag around that rather than the feature itself?

I could live with that.

src/Scientist/Internals/GCHelper.cs
```diff
@@ -0,0 +1,132 @@
...
+                            .GetTypeInfo()
+                            .GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.Public | BindingFlags.Static)
+                         ?? typeof(GC)
+                            .GetTypeInfo()
+                            .GetMethod("GetAllocatedBytesForCurrentThread",
+                            BindingFlags.NonPublic | BindingFlags.Static);
+            return () => (long)method.Invoke(null, null);
+#else
+            return () => 0L;
```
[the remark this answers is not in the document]
Lexcess: Ah I see, cache off the delegate like we do with .Net Core. I think at one point that wasn't possible (due to some additional GC.Collects), but you're right we can make this change. I'll try and find time to make the changes in the next day or so.

src/Scientist/Internals/GCHelper.cs
Lexcess: See below, I think this is addressed with your later comment.

src/Scientist/Internals/GCHelper.cs
Lexcess: Not seen that method before, I've swapped out the implementation for CreateDelegate.

src/Scientist/Internals/GCHelper.cs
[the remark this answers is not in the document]
Lexcess: So I've gone through and reworked the code. It does mean that different implementations are used to get byte counts (process wide versus current thread). However the perf tradeoff might be worth it. I've also bundled in the allocation quantum logic in the delegate.

Lexcess:
I've applied the PR feedback and got the tests passing. I haven't added a flag to enable opt in for NetFX's GC.Collect impl. Did you have a preference in mechanism for that? I wouldn't recommend config file dependencies on a library. Other than that there are numerous options we could #IfDef in a NetFX only overload or have a platform agnostic flag that only comes into play with NetFX. Any thoughts?

PR #124, opened 2019-02-25
Return results from experiment

paulbreen:
Scientist is designed to always return results from the control,  however during development I have found it helpful to return the result from the candidate rather than the control, i.e. during design and debugging an issue etc.

The following PR exposes a delegate that will can effectively ignore the control and return the result from the first candidate. It has similar affects to turning off Scientist i.e. no comparison reports are generated, no over head etc

```
using GitHub;

...

public void ScientistSetup(){
   Scientist.EnabledControl(() => !HostingEnvironment.IsDevelopmentEnvironment);
}

public bool UseExperimentResult(IUser user)
{
    return Scientist.Science<bool>("widget-permissions", experiment =>
    {
        experiment.Use(() => IsCollaborator(user)); // control
        experiment.Try(() => HasAccess(user)); // candidate
    }); // returns the candidate value
}
```

### DIFF
--- a/src/Scientist/Internals/ExperimentInstance.cs
+++ b/src/Scientist/Internals/ExperimentInstance.cs
@@ -22,6 +22,7 @@ namespace GitHub.Internals
         internal readonly Func<T, T, bool> Comparator;
         internal readonly Func<Task> BeforeRun;
         internal readonly Func<Task<bool>> Enabled;
+        internal readonly Func<Task<bool>> EnableControl;
         internal readonly Func<Task<bool>> RunIf;
         internal readonly IEnumerable<Func<T, T, Task<bool>>> Ignores;
         internal readonly Dictionary<string, dynamic> Contexts;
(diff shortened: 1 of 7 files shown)


src/Scientist/Internals/ExperimentInstance.cs
```diff
@@ -64,6 +66,11 @@ public async Task<T> Run()
                 return await Behaviors[0].Behavior();
             }
 
+            if (!await ShouldControlRun())
+            {
+                return await Behaviors[1].Behavior();
```
M-Zuber: This does not allow having multiple candidates, which makes sense in the context of the PR, but maybe this can be called out stronger by changing the name of `ShouldControlRun` to something like `ReturnFirstCandidate`?

paulbreen: Apologizes for the delay, 
I agree, i'll make the change.

paulbreen: renamed the property as requested

M-Zuber (approved):

PR #129, opened 2019-06-27
Set Enabled in StaticHelper to afect all instances

Sertrunku:
Background for this PR. Having a few experiments running without using the Static Helper, was trying to disable all experiments at once, by using the Enabled method. Turns out it wasn't possible.

To achieve this Had to create a implementation of Scientist and override the IsEnabledAsync.

This PR aims to allow the set of the Enabled func globally, for experiments using the helper and for the implementations. 
The IsEnabledAsync remains virtual so that we can still override the enabled behaviour for a specific one

### DIFF
--- a/src/Scientist/Scientist.cs
+++ b/src/Scientist/Scientist.cs
@@ -9,9 +9,8 @@ namespace GitHub
     /// </summary>
     public class Scientist : IScientist
     {
-        static readonly Task<bool> EnabledTask = Task.FromResult(true);
-        static readonly Lazy<Scientist> _sharedScientist = new Lazy<Scientist>(CreateSharedInstance);
-        static Func<Task<bool>> _enabled = () => EnabledTask;
+        static readonly Lazy<Scientist> SharedScientist = new Lazy<Scientist>(CreateSharedInstance);
+        static Func<Task<bool>> _enabled = () => Task.FromResult(true);
         static IResultPublisher _sharedPublisher = new InMemoryResultPublisher();
         readonly IResultPublisher _resultPublisher;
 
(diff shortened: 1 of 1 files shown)


M-Zuber:
I am not sure how the changes accomplish the stated goal, could you maybe provide some code samples to show the difference in use between the current code, and the code you are proposing?

Sertrunku:
Sure, so right now we have two ways of starting an Experiment, either by using the Static helper like
`Experiment<bool>(ExperimentName, experiment =>
            {
                experiment.Use(() => true);
                experiment.Try(() => false);
            });`
or by  creating a new Scientist object like this
`var scientist = new Scientist(resultPublisher);
            scientist.Experiment<bool>("", experiment =>
            {
                experiment.Use(() => true);
                experiment.Try(() => false);
            });`
Given this, if we have multiple experiments on both formats and we want to use the Enabled function to disable or enable all of the experiments the current implementation doesn't allow it. The current Enabled function only affects experiments created using the static helper.

My changes allow the Enabled to affect all of the esperiments regardless if they are being create using the static helper or the instance

M-Zuber:
One change that could be helpful in evaluating this would be to revert the naming change of `_sharedScientist` to `SharedScientist`

Do you think you could write a test that will only pass with this change?

Sertrunku:
Yeah sure, I'll write the test and update the PR.

Will revert the name change also

M-Zuber:
Thank you!
(also, you might to keep this in mind: https://github.com/scientistproject/Scientist.net/pull/108#issuecomment-390773865)

Sertrunku:
I do have a few more ideas to make it more IoC firendly, but I wanna do them in a different PR

Sertrunku:
@M-Zuber did you had time to review the latest changes?

M-Zuber (approved):

M-Zuber (approved):

PR #131, opened 2020-05-29
Resolve deadlock issue on async experiments when caller uses sync ove…

zizhong-zhang:
…r async

### DIFF
--- a/src/Scientist/Internals/ExperimentInstance.cs
+++ b/src/Scientist/Internals/ExperimentInstance.cs
@@ -58,15 +58,15 @@ namespace GitHub.Internals
         public async Task<T> Run()
         {
             // Determine if experiments should be run.
-            if (!await ShouldExperimentRun())
+            if (!await ShouldExperimentRun().ConfigureAwait(false))
             {
                 // Run the control behavior.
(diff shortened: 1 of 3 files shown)


zizhong-zhang:
We have some legacy mvc (netframework 4.6.1) which has lots of synchronous code.
We've found that when caller gets the Task.Result synchronously from Scientist.SicenceAsync<T>, context thread is blocked forever.

zizhong-zhang:
thank you for the instant response on this. this is truly amazing!
would you be okay to merge this in?

zizhong-zhang:
hi @paulbreen , thanks for help merging the PR in.
it would be fantastic if a new nuget release can be out so we will be able to consume it.

At the moment, we've forked your repo and publish a nuget from there however we know the best way to go long-term is to stick with the upstream.

Hope it does not take much effort for you to bump the version (e.g. 2.0.1)

Many thanks!

kmc059000:
We ran into a similar issue and in researching, we found that this PR had been merged almost a year ago, with no corresponding nuget package release. Would it be possible to deploy a new nuget package for this?

Issue #136, opened 2021-07-21
Version 3


JoshHiles:
Version 3 is about rewriting Scientist.net for .NET 6 & rejuvenating the project

Some actions to start with:
- New Version 3 branch (feature/version-3/net-6 maybe?)
- Issue for GitHub action to deploy to NuGet with 3.x.x-alpha
- Potential 1.x.x branch for maintance

paulbreen:
Great to see all the momentum with the new version. 
I don’t think this should be a direct port to .net 6. We should take our time to consider if there improvements or changes etc we’d like to make and what new things .net 6 can give us.

marblekirby:
Should we have a discussion about how we will approach the code rewrite to .net 6. Are we planning on porting everything from the current 2.x.x release to 3.0.0?

M-Zuber:
I've been thinking about this, and IMO for the 3.0.0 release we should strive to have no effective breaking changes behaviorally.
If we need/want to tweak the API that would be fine, but the user should not have to think about the changes much.
Afterwards we can start setting up an obsoletion policy and make deeper changes

JoshHiles:
Sounds good, I think we port everything across to the new solution with the upgraded .net and get everything setup, finish off all the fluff around docs, actions etc. Then take it from there perhaps?
We already have kinda a policy in place if you check #143

M-Zuber:
Yeah, sounds like what I had in mind 👍

marblekirby:
@JoshHiles I see alot of tickets open for the fluff around the docs/actions etc but what about the port to latest .net version?.

I am working through the wiki. I see alot of new actions created. What about getting started on the version migration?

JoshHiles:
Apolgies @marblekirby lots on at work!
I think it has already started with the new solution been created but as far as that im not sure anything else apart from the supporting community stuff has been done.
I'll make a start moving it over to the new solution when i have a spare minute

mwilson-atlassian:
👋 I followed the breadcrumbs from other issues to get here. Am I correct that the re-write ran out of steam and the project is unmaintained, at this point?

mwilson-atlassian:
@JoshHiles Great! I just got permission from my company to contribute, so I'll probably pull this into a spike I'm working on next week. Is there a discord or something like that for informal collab?

JoshHiles:
[...]

PR #152, opened 2021-12-05
Move code to new solution

JoshHiles:
As mentioned in #136 i've moved over the code to the new solution.

Unit tests pass locally & benchmarking runs too.

The only thing i haven't transfered across is any nuget, ci/cd stuff as of yet; this should be considered to be put into github actions


marblekirby:
Hi @JoshHiles, I have some time and will review this

PR #162, opened 2025-09-22
Newer Target Frameworks

dahlbyk:
Since v2.1 is adding `netstandard2.0` (f01d9e6d02f199f999405ed14c86192da7ae115d), it might as well add newer LTS targets.

I know `net6.0` is technically unsupported, but still a common target.

### DIFF
--- a/test/Scientist.Benchmark/Scientist.Benchmark.csproj
+++ b/test/Scientist.Benchmark/Scientist.Benchmark.csproj
@@ -4,7 +4,7 @@
     <IsPackable>false</IsPackable>
     <OutputType>Exe</OutputType>
     <RootNamespace>GitHub</RootNamespace>
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;net48</TargetFrameworks>
     <RestorePackagesWithLockFile>true</RestorePackagesWithLockFile>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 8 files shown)


test/Scientist.Benchmark/Scientist.Benchmark.csproj
```diff
@@ -4,7 +4,7 @@
     <IsPackable>false</IsPackable>
     <OutputType>Exe</OutputType>
     <RootNamespace>GitHub</RootNamespace>
-    <TargetFrameworks>net8.0</TargetFrameworks>
+    <TargetFrameworks>net8.0;net6.0;net48</TargetFrameworks>
```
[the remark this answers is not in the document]
dahlbyk: Executable can't target `netstandard`.

test/Scientist.Test/Scientist.Test.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
+    <TargetFrameworks>net8.0;net48</TargetFrameworks>
```
dahlbyk, line 4: Adding `net6.0` here yields an error:
```text
error : Microsoft.NET.Test.Sdk doesn't support net6.0 and has not been tested with it. Consider upgrading your TargetFramework to net8.0 or later.
You may also set <SuppressTfmSupportBuildErrors>true</SuppressTfmSupportBuildErrors> in the project file to ignore this warning and attempt to run in this unsupported configuration at your own risk.
```

If you do set `SuppressTfmSupportBuildErrors=true` there are other errors.

JoshHiles: Ah it will be because i have 8 & 9 installed!

| Since .NET 6 is out of support I don't think we need to got out of our way to test it going forward.

I completely agree

PR #166, opened 2025-11-14
Feature: experiment.EnsureControlRunsFirst()

### LINKED ISSUE #159: [Feature] Support a non-randomized execution order
While randomizing the execution order is probably the right thing to do for "most" cases, there are at least some cases whereby it's not desirable. 

We wanted to use this library for a "dry-run" - we needed to perform the "experiment" flow before performing the "control" flow (executing the control flow first would invalidate the results of the experiement flow, but not vice-versa).

**Describe the solution you'd like**
It would be useful to have an option that **disables** randomiziation (control/experiment) and instead executes the trials in specified order
* Experiement first, then control 
* Control first, then experiment

**Describe alternatives you've considered**
No alternatives exist that we can think of (outside of not using the library)

JoshHiles:
resolves #159

A new method `experiment.EnsureControlRunsFirst()` which if enabled will randomise the candidates first then add the control to start.

### DIFF
--- a/.github/workflows/codeql.yaml
+++ b/.github/workflows/codeql.yaml
@@ -1,0 +1,40 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ] 
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      packages: read
+      actions: read
+      contents: read
+
+    strategy:
+      fail-fast: false
+      matrix:
+         include:
+          - language: actions
+            build-mode: none
+          - language: csharp
+            build-mode: none
+
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Initialize CodeQL
+        uses: github/codeql-action/init@v3
+        with:
+          languages: ${{ matrix.language }}
+          build-mode: ${{ matrix.build-mode }}
+
+      - name: Perform CodeQL Analysis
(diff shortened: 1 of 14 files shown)


M-Zuber (approved):
The implementation is clean, and could go in as is

I am wondering though if this actually solves the issue raised. Since sometimes I can want the experiments to run, and only then run the control. For example, if each experimental behaviour does clean up and just reports on results, then I want the control to run last, without doing any cleanup.
 Which then leads us to the bikeshedding of allowing the user to optionally pass in an ordering with each behaviour, or a set number of use cases controlled by settings.

JoshHiles:
Legend, very nice catch, i have been very narrow focused on just the issued raised!
Damn yeh a few options then i guess

## Pass order to `Try` & `Catch`
```csharp
experiment.Use(ControlMethod, order: 0);
experiment.Try(CandidateMethod, order: 1);
```
### Pros
- Full flexibility
### Cons
- I think its a bit of UX nightmare
- Have to deal with order clashes
- If you set the order on one what about the order of the rest? Lots of edge cases

## Enum defined order method
```csharp
experiment.WithExecutionOrder(OrderExecution.ControlFirst)
```

### Pros
- Clearly defined API 
- Easy to add new orders
### Cons
- Less flexible

## User defined ordering
```csharp
experiment.UseCustomOrder(behaviours  => CustomMethod(behaviours))
```

### Pros
- Most flexible in that user gets complete control
- We can expose some methods to use
- We don't have to catch edge cases and custom logic
### Cons
- ?

Each way has its own trade offs and advantages, i hesitate to implement them all but perhaps that might be the way?

PR #168, opened 2025-11-21
feat: Asynchronous experiment cancellation

### LINKED ISSUE #165: [Feature] Support cancellable experiments
**Is your feature request related to a problem? Please describe.**
While using async experiments, I found myself desiring to cancel experiments mid experiment without it contributing to mismatch or error

**Describe the solution you'd like**
I would like the ability to pass through a Cancellation token into the experiment such that when `IsCancellationRequested` is true, then the experiment is cancelled safely, and potentially the fact that the experiment has been cancelled is published. I would imagine this is either passed through as the last optional argument of `Scientist.ScienceAsync` or added a new method added to IExperimentAsync like `CancellableBy`. I imagine 

**Describe alternatives you've considered**
Pass through the cancellation token to the control and candidate but that could result in incorrect reporting, then require modification to the result publisher

I am happy to contribute to the solution.

Thanks

JoshHiles:
this closes #165

New `experiment.WithCancellationToken()` & overloads for the async `Try` & `Catch`

### DIFF
--- a/src/Scientist/Internals/Experiment.cs
+++ b/src/Scientist/Internals/Experiment.cs
@@ -1,6 +1,7 @@
 ﻿using Github.Ordering;
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 
 namespace GitHub.Internals
@@ -16,9 +17,9 @@ namespace GitHub.Internals
         private string _name;
         private int _concurrentTasks;
 
-        private Func<Task<T>> _control;
+        private Candidate<T> _control;
 
-        private readonly Dictionary<string, Func<Task<T>>> _candidates;
+        private readonly Dictionary<string, Candidate<T>> _candidates;
         private Func<T, TClean> _cleaner;
         private Func<T, T, bool> _comparison = DefaultComparison;
         private Func<Task> _beforeRun;
@@ -31,6 +32,9 @@ namespace GitHub.Internals
 
         private CustomOrderer<T> _customOrderer = behaviors => Task.FromResult(Ordering.Random(behaviors));
 
+        private CancellationToken _cancellationToken = default;
+
+
         public Experiment(string name, Func<Task<bool>> enabled, int concurrentTasks, IResultPublisher resultPublisher)
         {
             if (concurrentTasks <= 0)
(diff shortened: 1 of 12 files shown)


JoshHiles:
Tests were failing on [a21eb17](https://github.com/scientistproject/Scientist.net/pull/168/commits/a21eb17befca55a9bde70543f3695d8dcc2fcfc4) as the `EnsureNullGuardIsWorking` tests don't run in `DEBUG` mode, I'm wondering if we should remove this? 

Anyways the null ref ex throws earlier as we build the Behaviours from the Candidate which causes the null ref ex to throw earlier. I think this is correct way to fix but open to feedback

src/Scientist/Internals/Experiment.cs
```diff
@@ -60,20 +64,34 @@ public void RunIf(Func<bool> block) =>
         public void Thrown(Action<Operation, Exception> block) =>
             _thrown = block;
 
-        public void Use(Func<Task<T>> control) =>
-            _control = control;
+        public void Use(Func<Task<T>> control, CancellationToken cancellationToken = default)
+        {
+            var isasd = cancellationToken == default
```
M-Zuber: where is this variable used?

M-Zuber (dismissed):
LGTM
there is that one variable that looks like it should have been removed

src/Scientist/Internals/Experiment.cs
JoshHiles: Nice spot! I've removed the variable.

JoshHiles:
> there is that one variable that looks like it should have been removed

I've double checked the project build settings and we do have level 4 warnings & we don't have `CS0168` excluded.
Just wondering if there's a way to catch this next time in VS?
Is this because we support net451 ?

M-Zuber:
> > there is that one variable that looks like it should have been removed
> 
> I've double checked the project build settings and we do have level 4 warnings & we don't have `CS0168` excluded. Just wondering if there's a way to catch this next time in VS? Is this because we support net451 ?

TBH I don't know
(and personally I am not _that_ concerned about this particular error...)

M-Zuber (approved):

JoshHiles:
Groovy!
Apologies in advance, their will be another PR through for the release notes at somepoint

